PR #2, opened 2020-03-24
Add extension array for N-dimensional tensors

BryanCutler:
This adds a `TensorArray` as a Pandas extension array so that it is possible to work with tensor values inside of a Pandas DataFrame.

The `TensorArray` can be constructed with a single numpy.ndarray where the outer dimension represents the number of ndarrays, or tensors, in the column. Alternatively, a `TensorArray` can be constructed with a list of ndarrays of individual tensor values.


BryanCutler:
@frreiss this is a pretty rough wip, but some things are working so we might be able start testing some things with it.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __getitem__(self, item) -> "TensorArray":
+        """
+        See docstring in `Extension   Array` class in `pandas/core/arrays/base.py`
+        for information about this method.
+        """
+        # TODO pandas converts series with np.asarray, then applied a function e.g. map_infer(array, is_float) to format strings etc.
+        # If single element return as np.ndarray, if slice return TensorArray of slice
+        if isinstance(item, int):
+            return self._tensor[item]
```
BryanCutler: This allows a `TensorArray` to be used in `np.asarray` and result in a `ndarray` with the correct dtype. It's needed because I ran into problems otherwise when used as a column inside a DataFrame.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+from memoized_property import memoized_property
+
+# Internal imports
+import text_extensions_for_pandas.util as util
+
+
+@pd.api.extensions.register_extension_dtype
+class TensorType(pd.api.extensions.ExtensionDtype):
+    """
+    Panda datatype for a span that represents a range of characters within a
```
frreiss: This docstring needs to be updated.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+    target string.
+    """
+
+    @property
+    def type(self):
+        # The type for a single row of a column of type CharSpan
```
frreiss: Comment needs to be updated and perhaps converted to a docstring.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+        for information about this method.
+        """
+        return TensorArray
+
+
+class TensorArray(pd.api.extensions.ExtensionArray):
+    """
+    A Pandas `ExtensionArray` that represents a column of `numpy.ndarray`s,
+    or tensors, where the outer dimension is the count of tensors in the column.
+    """
```
frreiss: It would be good to note here that all the tensors in a column must be of the same shape.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __init__(self, values: Union[np.ndarray, List[np.ndarray]],
```
frreiss: I've started using `Sequence` instead of `List` in public APIs' type hints. I see that you found a place where my code was still using `List` and copied that :-). I'll fix the constructors for `TokenSpan` and `CharSpan` in a separate commit. Can you change the `List[np.ndarray]` to `Sequence[np.ndarray]` here?

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+                 make_contiguous: bool = True):
+        """
+        :param values: A `numpy.ndarray` or list of `numpy.ndarray`s
+        :param make_contiguous: force values to be contiguous in memory
+        """
+        if isinstance(values, list):
```
frreiss: `list` --> `Sequence`

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+        elif isinstance(values, np.ndarray):
+            self._tensor = values
+        else:
+            raise TypeError("Expected a numpy.ndarray or list of numpy.ndarray")
+        
+        if not self._tensor.flags.c_contiguous:
+            if make_contiguous:
+                self._tensor = np.ascontiguousarray(self._tensor)
+            else:
+                raise ValueError("Input ndarray is not contiguous")
```
frreiss: I don't think this branch is an error condition. The user has set `make_contiguous` to `False`, which means that the internal representation can be either contiguous or not.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @classmethod
+    def _concat_same_type(
+        cls, to_concat: Sequence["TensorArray"]
+    ) -> "TensorArray":
+        """
+        See docstring in `ExtensionArray` class in `pandas/core/arrays/base.py`
+        for information about this method.
+        """
+        return TensorArray([a._tensor for a in to_concat])
```
frreiss: Can drop the list comprehension here once the constructor accepts `Sequence` as input.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+        # TODO any or all values in row nan?
+        return np.any(np.isnan(self._tensor), axis=1)
+
+    def copy(self) -> "TensorArray":
+        """
+        See docstring in `ExtensionArray` class in `pandas/core/arrays/base.py`
+        for information about this method.
+        """
+        ret = TensorArray(
+            self._tensor,
```
frreiss: To be on the safe side, you probably want to tell the ndarray to copy itself here.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def take(
+        self, indices: Sequence[int], allow_fill: bool = False,
+        fill_value: Any = None
+    ) -> "TensorArray":
+        """
+        See docstring in `ExtensionArray` class in `pandas/core/arrays/base.py`
+        for information about this method.
+        """
+        # TODO if allow_fill:
```
frreiss: For now, can you make this method throw `NotImplementedException` if `allow_fill` is `True`?

text_extensions_for_pandas/array/tensor.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __len__(self) -> int:
+        return len(self._tensor)
+
+    def __eq__(self, other):
+        return TensorArray(self._tensor == other._tensor)
+
+    def __lt__(self, other):
+        """
+        Pandas-style array/series comparison function.
```
frreiss: Docstring needs updating, or you can just remove it.

text_extensions_for_pandas/array/tensor.py
frreiss: Yes, `__getitem__` needs to return the "scalar" type when the index is an `int` and the "array" type when the index is a slice/list/iterable. The API docs don't make this point terribly clear.

text_extensions_for_pandas/array/tensor_test.py
```diff
@@ -0,0 +1,96 @@
...
+        s2 = TensorArray(x)
+        self.assertTrue(np.all(s1 == s2))
+
+        s2 = TensorArray(x * 2)
+        self.assertTrue(np.all(s1 < s2))
+        self.assertFalse(np.any(s1 > s2))
+
+    def test_repr(self):
+        x = np.array([[1, 2], [3, 4], [5, 6]])
+        expected = """\
```
frreiss: You can use `textwrap.dedent` to make these string literals line up with the code. See https://github.com/frreiss/text-extensions-for-pandas/blob/667ad70ed28440b76497f72df9f654c08d28e9dd/text_extensions_for_pandas/spanner/test_join.py#L69

text_extensions_for_pandas/array/tensor.py
BryanCutler: Ah, ok good then!

text_extensions_for_pandas/array/tensor.py
BryanCutler: Sure, good idea

text_extensions_for_pandas/array/tensor.py
BryanCutler: oh right, I think this was held over from the arrow extension, where it had to be contiguous

text_extensions_for_pandas/array/tensor.py
BryanCutler: sure

BryanCutler:
Thanks @frreiss , I'll work on an update

text_extensions_for_pandas/array/tensor.py
BryanCutler: I'll remove, I'm going to try a different way to define the operators

text_extensions_for_pandas/array/tensor_test.py
BryanCutler: worked great, thanks!

frreiss:
Thanks! Merging.

PR #3, opened 2020-03-30
Add mixin for defining operations based on internal ndarray

BryanCutler:
This uses a Pandas extension base class to define comparison and arithmetic operators that are based on the internal ndarray.


BryanCutler:
@frreiss this seems like a better way to define operators since they are all based on those from ndarray

BryanCutler:
Thanks @frreiss !

>Should something similar be done for any of the other ExtensionArrays in this project?

Maybe, it's only useful if you can generalize all operators in the same way. I’ll take a look.

PR #8, opened 2020-04-22
Add wrapper function for BERT tokenization

BryanCutler:
Add function `make_bert_tokens` to wrap a huggingface `PreTrainedTokenizerFast` tokenizer and return result as a Pandas DataFrame that includes all data from the tokenizer with character span, and token span columns.

### DIFF
--- a/text_extensions_for_pandas/io/tokenization.py
+++ b/text_extensions_for_pandas/io/tokenization.py
@@ -1,0 +1,97 @@
+#
+#  Copyright (c) 2020 IBM Corp.
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+################################################################################
+# tokenize.py
+#
+# Functions for tokenization of text
+
+import numpy as np
+import pandas as pd
+
+from text_extensions_for_pandas.array import (
+    CharSpanArray,
+    TokenSpanArray,
+)
+
+
+def make_bert_tokens(target_text: str, tokenizer) -> pd.DataFrame:
+    """
+    Tokenize the indicated text for BERT embeddings and return a DataFrame
+    with one row per token.
+
+    :param: target_text: string to tokenize
+    :param: tokenizer: A tokenizer that is a subclass of huggingface transformers
+                       PreTrainingTokenizerFast which supports `encode_plus` with
+                       return_offsets_mapping=True.
+
+    :returns: A `pd.DataFrame` with the following columns:
+     * "id": unique integer ID for each token
+     * "char_span": span of the token with character offsets
+     * "token_span": span of the token with token offsets
+     * "input_id": integer ID suitable for input to a BERT embedding model
+     * "token_type_id": list of token type ids to be fed to a model
+     * "attention_mask": list of indices specifying which tokens should be
+                         attended to by the model
+     * "special_tokens_mask": `True` if the token is a zero-length special token
+       such as "start of document"
+    """
+    from transformers.tokenization_utils import PreTrainedTokenizerFast
+    if not isinstance(tokenizer, PreTrainedTokenizerFast):
+        raise TypeError("Tokenizer must be an instance of "
+                        "transformers.PreTrainedTokenizerFast that supports "
+                        "encode_plus with return_offsets_mapping=True.")
+    tokenized_result = tokenizer.encode_plus(target_text,
+                                             return_special_tokens_mask=True,
+                                             return_offsets_mapping=True)
+
+    # Get offset mapping from tokenizer
+    offsets = tokenized_result["offset_mapping"]
+
+    # Init any special tokens at beginning
+    i = 0
+    while offsets[i] is None:
+        offsets[i] = (0, 0)
+        i += 1
+
+    # Make a DataFrame to unzip (begin, end) offsets
+    offset_df = pd.DataFrame(offsets, columns=["begin", "end"])
+
+    # Convert special tokens mask to boolean
+    special_tokens_mask = \
+        pd.Series(tokenized_result["special_tokens_mask"]).astype('bool')
+
+    # Fill remaining special tokens to zero-length spans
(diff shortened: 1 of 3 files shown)


text_extensions_for_pandas/transformers/tokenization.py
```diff
@@ -0,0 +1,90 @@
...
+                                             return_offsets_mapping=True)
+    df = pd.DataFrame.from_dict(tokenized_result)
+    # Get offset mapping from tokenizer
+    offsets = tokenized_result["offset_mapping"]
+
+    # Turn special tokens into zero-length spans
+    begins = np.zeros(len(df.index), dtype=np.int64)
+    ends = np.zeros(len(df.index), dtype=np.int64)
+    prev_begin = 0
+    for i in range(len(df.index)):
```
frreiss, line 64: Can you vectorize this loop?
You should be able to implement this transformation as follows: 
1. Correct the `begin` values for any special tokens at the very beginning of the text (either set them to 0, or set them to the begin offset of the first non-special token).
2. Use [`pandas.DataFrame.fillna`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.fillna.html#pandas.DataFrame.fillna) with the `'ffill'` imputation method to fill in the missing `begin`.
3. Replace missing `end` values with the corresponding value from the corrected `begin` series.

text_extensions_for_pandas/transformers/tokenization.py
```diff
@@ -0,0 +1,90 @@
...
+            begins[i] = prev_begin
+            ends[i] = prev_begin
+        else:
+            begins[i], ends[i] = offsets[i]
+            prev_begin = begins[i]
+
+    char_spans = CharSpanArray(target_text, begins, ends)
+    df["char_span"] = char_spans
+    df["token_span"] = TokenSpanArray(char_spans,
+                                      list(range(len(char_spans))),
```
frreiss, line 76: These initializer should be using `np.linspace` to build up those begin and end lists.

text_extensions_for_pandas/transformers/test_tokenization.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertEqual(span.begin, span.end)  # docend special token
+        expected = ["the", "quick", "brown", "fox", "over", "a", "bog"]
+        tokens = [span.covered_text for span in char_span]
+        for e in expected:
+            self.assertTrue(e in tokens)
+
+        # verify token span array
+        self.assertTrue('token_span' in bert_toks.columns)
+        token_span = bert_toks['token_span']
+        span = char_span.iloc[0]
```
frreiss, line 70: `char_span` should be `token_span`

text_extensions_for_pandas/transformers/test_tokenization.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertEqual(span.begin, span.end)  # docstart special token
+        span = char_span.iloc[-1]
```
frreiss, line 72: `char_span` should be `token_span`

text_extensions_for_pandas/transformers/test_tokenization.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertEqual(span.begin, span.end)  # docend special token
```
frreiss, line 73: You should check whether the `begin` and `end` offsets of the token spans align properly with the corresponding character-based spans.

text_extensions_for_pandas/transformers/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+################################################################################
+# transformers module
+#
+# Functions in text_extensions_for_pandas to tokenize text and put results in
+# a Pandas DataFrame.
```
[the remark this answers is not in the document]
frreiss: Alternately, you could do away with this `__init__.py` file entirely and just add `transformers.py` directly to the directory `text_extensions_for_pandas/io`

text_extensions_for_pandas/transformers/test_tokenization.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertEqual(special_tokens_mask.iloc[-1], True)
+
+        # verify char span array
+        self.assertTrue('char_span' in bert_toks.columns)
+        char_span = bert_toks['char_span']
+        span = char_span.iloc[0]
+        self.assertEqual(span.begin, span.end)  # docstart special token
+        span = char_span.iloc[-1]
+        self.assertEqual(span.begin, span.end)  # docend special token
+        expected = ["the", "quick", "brown", "fox", "over", "a", "bog"]
```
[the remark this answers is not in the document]
BryanCutler: I wasn't sure if those would be tokenized consistently or it might depend on some setting since they aren't words. Should I check the output as is or use real words instead?

BryanCutler:
Thanks @frreiss , all feedback sounds good. I didn't put it in the io directory because it is only tokenization, but it makes sense to be there since it's related to the io functionality, so I will move.

text_extensions_for_pandas/transformers/test_tokenization.py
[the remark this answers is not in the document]
BryanCutler: I used real words, we can test fancier things later

text_extensions_for_pandas/transformers/test_tokenization.py
BryanCutler: I added a check that the `covered_text` is the same, is that what you mean?

text_extensions_for_pandas/io/tokenization.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # Make a DataFrame to unzip (begin, end) offsets
+    offset_df = pd.DataFrame(offsets, columns=["begin", "end"])
+
+    # Convert special tokens mask to boolean
+    special_tokens_mask = \
+        pd.Series(tokenized_result["special_tokens_mask"]).astype('bool')
+
+    # Fill remaining special tokens to zero-length spans
+    ends = offset_df['end'].fillna(method='ffill').astype('int32')
```
BryanCutler, line 77: I believe it's correct to fflill the ends and then copy over to the begins. If done the other way to ffill the begins, then the last real token will have (45, 46), and end doc token will have (45, 45) offsets, for example, which seems wrong.

BryanCutler:
@frreiss I believe I made all the updates. Please take another look when you can, thanks!

text_extensions_for_pandas/io/tokenization.py
frreiss: Makes sense.

BryanCutler:
Thanks @frreiss !

PR #9, opened 2020-04-24
Add arrow extension type for TensorArray IO

BryanCutler:
Add Arrow extension for TensorArray and conversion function to support IO from a DataFrame. Currently, Feather format is working but Parquet has an error when reading back.

### DIFF
--- a/text_extensions_for_pandas/array/tensor.py
+++ b/text_extensions_for_pandas/array/tensor.py
@@ -70,6 +70,11 @@ class TensorType(pd.api.extensions.ExtensionDtype):
         """
         return TensorArray
 
+    def __from_arrow__(self, extension_array):
+        from text_extensions_for_pandas.array.arrow_compat import ArrowTensorArray
+        values = ArrowTensorArray.to_numpy(extension_array)
+        return TensorArray(values)
+
 
 class TensorOpsMixin(pd.api.extensions.ExtensionScalarOpsMixin):
     """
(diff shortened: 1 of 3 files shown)


BryanCutler:
@frreiss this is working for me using Feather format, but Parquet seems to get an error when reading back in that I'll have to look.into. I think Feather should work for your demo if you want to give this a shot. I am using the newly released pyarrow 0.17.0, and haven't tried against 0.16.0 yet.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -70,6 +70,11 @@ def construct_array_type(cls):
         """
         return TensorArray
 
+    def __from_arrow__(self, extension_array):
+        from text_extensions_for_pandas.array.arrow_compat import ArrowTensorArray
```
BryanCutler, line 74: I put this in a new module because I'm not sure if pyarrow is a hard dependency on pandas, we can move it if you like

frreiss: Not having a hard dependency on Arrow makes sense, at least for now. We may want to revisit that once we get around to having a (presumably Arrow-backed) Pandas extension type for variable-length tensors.

BryanCutler:
Thanks @frreiss !  I sent a message to the arrow dev list about Parquet support and it seems like it _should_ work, so I'll keep looking at it.

PR #10, opened 2020-04-29
Add function spans_to_iob to convert token spans to iob entities

BryanCutler:


BryanCutler:
@frreiss not sure if I got the logic right on this one, let me know if I'm totally off. Thanks!

text_extensions_for_pandas/io/conll.py
```diff
@@ -462,6 +462,46 @@ def iob_to_spans(
...
+    Inside–Outside–Beginning (IOB) format.
+    :param span_frame: DataFrame with a `TokenSpanArray` column.
+    :param token_span_col_name: Name of a column in `token_features` that contains the
+     IOB tags as strings, "I", "O", or "B".
+    :return: A `pd.DataFrame` with the following columns:
+    * `ent_iob`: entity type in IOB format.
+    """
+    token_spans = span_frame[token_span_col_name]
+
+    # Define the IOB categorical type with 'B'==0, 'I'==1, 'O'==2
```
frreiss, line 480: You probably want `O` to be zero here.

text_extensions_for_pandas/io/conll.py
```diff
@@ -462,6 +462,46 @@ def iob_to_spans(
...
+    dtype = pd.CategoricalDtype(['B', 'I', 'O'], ordered=False)
+
+    # Handle an empty token span array
+    if len(token_spans) == 0:
+        return pd.DataFrame({'ent_iob': pd.Categorical([], dtype)})
+
+    # Initialize an IOB series with all 'O' entities
+    iob_data = np.ones(len(token_spans[0].tokens), dtype='int64') * 2
```
frreiss, line 488: If you use zero for `O`, then `np.ones(custom shape) * 2` becomes `np.zeros_like` here

frreiss:
This looks good. Merging.

I did some additional poking around, and it looks like the format we've been calling "IOB" is actually supposed to be called "IOB2". See [wikipedia](https://en.wikipedia.org/wiki/Inside%E2%80%93outside%E2%80%93beginning_(tagging)). Would you mind doing a follow-on pass to change references to "IOB" in our code and notebooks to "IOB2" and include pointers to that Wikipedia page?

text_extensions_for_pandas/io/conll.py
BryanCutler: I was trying to follow the category codes that I saw somewhere else, which looked like 'O' == 2. Would it help to define these categories as a list somewhere?

BryanCutler:
Thanks @frreiss !  Sure I will do a followup to change references to IOB2.

text_extensions_for_pandas/io/conll.py
frreiss: You probably saw Pandas assigning arbitrary numbers to the distinct values. If you're coding the I, O, and B manually, then it's better for O to be zero because that tag means "zero entities present at this position".

PR #11, opened 2020-05-05
Change IOB references to indicate IOB2 format

BryanCutler:
The IOB2 format is what is being used, change references to indicate that along with links to wikipedia entry https://en.wikipedia.org/wiki/Inside%E2%80%93outside%E2%80%93beginning_(tagging).


BryanCutler:
@frreiss I just changed references to IOB and not variable names, e.g. `ent_iob`. Let me know if you also want to change those.

PR #12, opened 2020-05-07
Fix dependencies for pip install

frreiss:
Upon further reading, it looks like `pip install` ignores `requirements.txt` files in the package and only pays attention to the `install_requires` argument of `setuptools.setup`. This commit moves all the dependencies information from the former to the latter location. This change should hopefully fix issues with dependencies not being installed alongside Text Extensions for Pandas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ To build a pip package from your local copy:
 
 To build and install a pip package from the head of the master branch:
 ```
-pip install git+https://github.com/frreiss/text-extensions-for-pandas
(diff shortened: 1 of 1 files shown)


BryanCutler:
@frreiss , I think most projects read the requirements.txt and add to `install_requires` that way. I could fix that up if you like.

frreiss:
Sure, if you could set that up, it would be great, @BryanCutler !

frreiss:
Dropping this change set in favor of #13

PR #13, opened 2020-05-08
Add dependent packages from requirements to setup.py

BryanCutler:
This adds dependent packages from requirements.txt to setup.py so that when the user installs, the required dependencies will be installed also.


frreiss:
LGTM. Thanks, Bryan!

PR #14, opened 2020-05-12
Implement Arrow conversions for char and token span arrays

BryanCutler:
Add Arrow conversions to/from char and token span arrays to enable saving DataFrames with these columns to feather files. This first implementation is not optimized for storage space and will save target text in the field metadata, possibly duplicated. Also, TokenSpanArrays will save a duplicate of the dependent CharSpanArrays in the same struct column. These could be improved in followup changes.

### DIFF
--- a/text_extensions_for_pandas/array/arrow_conversion.py
+++ b/text_extensions_for_pandas/array/arrow_conversion.py
@@ -14,7 +14,7 @@
 #
 
 #
-# arrow_compat.py
+# arrow_conversion.py
 #
 # Part of text_extensions_for_pandas
 #
@@ -64,7 +64,7 @@ class ArrowCharSpanType(pa.PyExtensionType):
 
 class ArrowTokenSpanType(pa.PyExtensionType):
     """
-    PyArrow extension type definition for conversions to/from CharSpan columns
+    PyArrow extension type definition for conversions to/from TokenSpan columns
     """
 
     TARGET_TEXT_KEY = ArrowCharSpanType.TARGET_TEXT_KEY
@@ -73,7 +73,7 @@ class ArrowTokenSpanType(pa.PyExtensionType):
 
     def __init__(self, index_dtype, target_text):
         """
-        Create an instance of a CharSpan data type with given index type and
+        Create an instance of a TokenSpan data type with given index type and
         target text that will be stored in Field metadata.
 
         :param index_dtype:
--- a/text_extensions_for_pandas/array/arrow_conversion.py
+++ b/text_extensions_for_pandas/array/arrow_conversion.py
@@ -71,7 +71,7 @@ class ArrowTokenSpanType(pa.PyExtensionType):
     BEGINS_NAME = "token_begins"
     ENDS_NAME = "token_ends"
 
-    def __init__(self, index_dtype, target_text):
+    def __init__(self, index_dtype, target_text, num_char_span_splits):
         """
         Create an instance of a TokenSpan data type with given index type and
         target text that will be stored in Field metadata.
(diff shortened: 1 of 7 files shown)


text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -0,0 +1,327 @@
...
+
+        pa.PyExtensionType.__init__(self, pa.struct(fields))
+
+    def __reduce__(self):
+        index_dtype = self.storage_type[self.BEGINS_NAME].type
+        metadata = self.storage_type[ArrowCharSpanType.BEGINS_NAME].metadata
+        return ArrowTokenSpanType, (index_dtype, metadata)
+
+
+def char_span_to_arrow(char_span):
```
frreiss, line 102: Missing type hints here. Should be `def char_span_to_arrow(char_span) -> pyarrow.ExtensionArray:`

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -0,0 +1,327 @@
...
+
+    typ = ArrowCharSpanType(begins_array.type, char_span.target_text)
+    fields = list(typ.storage_type)
+
+    storage = pa.StructArray.from_arrays([begins_array, ends_array], fields=fields)
+
+    return pa.ExtensionArray.from_storage(typ, storage)
+
+
+def arrow_to_char_span(extension_array):
```
frreiss, line 124: Missing type hints.

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -0,0 +1,327 @@
...
+            raise ValueError("Only pyarrow.Array with a single chunk is supported")
+        extension_array = extension_array.chunk(0)
+
+    assert pa.types.is_struct(extension_array.storage.type)
+
+    # Get target text from the begins field metadata and decode string
+    metadata = extension_array.storage.type[ArrowCharSpanType.BEGINS_NAME].metadata
+    target_text = metadata[ArrowCharSpanType.TARGET_TEXT_KEY]
+    if isinstance(target_text, bytes):
+        target_text = target_text.decode()
```
frreiss, line 143: Should this line specify a specific encoding?

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -0,0 +1,327 @@
...
+    ends_array = extension_array.storage.field(ArrowCharSpanType.ENDS_NAME)
+
+    # Zero-copy convert arrays to numpy
+    begins = begins_array.to_numpy()
+    ends = ends_array.to_numpy()
+
+    return CharSpanArray(target_text, begins, ends)
+
+
+def token_span_to_arrow(token_span):
```
frreiss, line 156: Missing type hints.

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -0,0 +1,327 @@
...
+    extension array can be serialized and transferred with standard
+    Arrow protocols.
+
+    :param token_span: A TokenSpanArray to be converted
+    :return: pyarrow.ExtensionArray containing TokenSpan data
+    """
+    # Create array for begins, ends
+    token_begins_array = pa.array(token_span.begin_token)
+    token_ends_array = pa.array(token_span.end_token)
+    char_begins_array = pa.array(token_span.begin)
```
frreiss, line 169: This line of code will store the character-based begin offset of each *span* in `token_span`. You should be storing the character-based begin offset of each token in the underlying tokens behind `token_span`.

text_extensions_for_pandas/array/test_token_span.py
```diff
@@ -292,5 +294,18 @@ def test_as_frame(self):
         self.assertEqual(len(df), len(arr))
 
 
+class TokenSpanArrayIOTests(ArrayTestBase):
+
+    def test_feather(self):
+        arr = self._make_spans_of_tokens()
```
frreiss, line 300: You need to create a second test case that creates `TokenSpan`s that cover more than one token per span. It would also be a good idea to test that zero-length spans are correctly serialized/deserialized.

text_extensions_for_pandas/array/arrow_conversion.py
BryanCutler: Ahh ok, got it

text_extensions_for_pandas/array/arrow_conversion.py
BryanCutler: I don't think it's necessary, pyarrow does the same I believe

BryanCutler:
@frreiss I think this is working now, and hopefully I'm testing against all possible cases now. I had a crazy bug that took me a while to figure out. It was due to copy/paste a line with a comma at the end in the `ArrowTokenSpanType` class, which resulted in a SIGSEGV and no other stacktrace! Took me a while to figure out what was wrong  :sob:

PR #17, opened 2020-05-18
Fix for TensorArray slices that are empty

### LINKED ISSUE #16: BUG: Indexing with an array of False values ==> crash
Code to reproduce:

```python
import text_extensions_for_pandas as tp
import pandas as pd
df = pd.DataFrame({
    "col1": tp.TensorArray([[1, 2], [3, 4]]) 
})
df[[False, False]]
```

Expected result: Empty dataframe.

Actual result:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-763acf5b214d> in <module>
      2     "col1": tp.TensorArray([[1, 2], [3, 4]])
      3 })
----> 4 df[[False, False]]

~/opt/miniconda3/envs/pd/lib/python3.7/site-packages/pandas/core/frame.py in __getitem__(self, key)
   2789         # Do we have a (boolean) 1d indexer?
   2790         if com.is_bool_indexer(key):
-> 2791             return self._getitem_bool_array(key)
   2792 
   2793         # We are left with two options: a single key, and a collection of keys,

[... 7 levels of Pandas internals stack trace ...]

~/pd/fred-tep/text_extensions_for_pandas/array/tensor.py in take(self, indices, allow_fill, fill_value)
    169                     # of each row.
    170                     values[i] = fill_value
--> 171         return TensorArray(values)
    172 
    173     @property

~/pd/fred-tep/text_extensions_for_pandas/array/tensor.py in __init__(self, values, make_contiguous)
    111         """
    112         if isinstance(values, Iterable):
--> 113             self._tensor = np.stack(values, axis=0)
    114         elif isinstance(values, np.ndarray):
    115             self._tensor = values

<__array_function__ internals> in stack(*args, **kwargs)

~/opt/miniconda3/envs/pd/lib/python3.7/site-packages/numpy/core/shape_base.py in stack(arrays, axis, out)
    420     arrays = [asanyarray(arr) for arr in arrays]
    421     if not arrays:
--> 422         raise ValueError('need at least one array to stack')
    423 
    424     shapes = {arr.shape for arr in arrays}

ValueError: need at least one array to stack
```

BryanCutler:
TensorArray slices that are empty have to be handled differently to return an empty np.array

Fixes #16

### DIFF
--- a/text_extensions_for_pandas/array/test_tensor.py
+++ b/text_extensions_for_pandas/array/test_tensor.py
@@ -207,6 +207,25 @@ class TestTensor(unittest.TestCase):
         expected = np.array([[3, 4], [5, 6]])
         npt.assert_array_equal(expected, result)
 
+    def test_bool_indexing(self):
+        s = TensorArray([[1, 2], [3, 4]])
+
+        result = s[[True, True]]
+        expected = np.array([[1, 2], [3, 4]])
+        npt.assert_array_equal(result, expected)
+
+        result = s[[True, False]]
+        expected = np.array([[1, 2]])
+        npt.assert_array_equal(result, expected)
+
+        result = s[[False, True]]
+        expected = np.array([[3, 4]])
+        npt.assert_array_equal(result, expected)
+
+        result = s[[False, False]]
+        expected = np.empty((0, 2))
+        npt.assert_array_equal(result, expected)
+
     def test_asarray(self):
         x = np.array([[1, 2], [3, 4], [5, 6]])
         s = TensorArray(x)
--- a/text_extensions_for_pandas/array/test_tensor.py
+++ b/text_extensions_for_pandas/array/test_tensor.py
@@ -42,6 +42,12 @@ class TestTensor(unittest.TestCase):
         s = TensorArray(x)
         self.assertEqual(len(s), 5)
 
+        x = np.empty((0, 2))
+        s = TensorArray(x)
+        self.assertEqual(len(s), 0)
+        s = TensorArray([])
+        self.assertEqual(len(s), 0)
+
         x = [np.ones([2, 3]), np.ones([3, 2])]
         with self.assertRaises(ValueError):
             TensorArray(x)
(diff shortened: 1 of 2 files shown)


text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -207,6 +207,25 @@ def test_slice(self):
         expected = np.array([[3, 4], [5, 6]])
         npt.assert_array_equal(expected, result)
 
+    def test_bool_indexing(self):
+        s = TensorArray([[1, 2], [3, 4]])
+
+        result = s[[True, True]]
+        expected = np.array([[1, 2], [3, 4]])
```
frreiss, line 214: Boolean indexing should return a TensorArray, just like indexing with a slice does.

text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -207,6 +207,25 @@ def test_slice(self):
...
+        npt.assert_array_equal(result, expected)
+
+        result = s[[False, True]]
+        expected = np.array([[3, 4]])
+        npt.assert_array_equal(result, expected)
+
+        result = s[[False, False]]
+        expected = np.empty((0, 2))
+        npt.assert_array_equal(result, expected)
+
```
frreiss, line 228: You need to test whether indexing with Boolean masks into `pd.Series` of `TensorArray` works.

You also need to test whether indexing with a Boolean mask into a DataFrame works. See the example code in #16

text_extensions_for_pandas/array/test_tensor.py
BryanCutler: it does, but I should add a check for the type as well
```
In [43]: arr = tp.TensorArray([[1, 2], [3, 4]])                                                                              

In [44]: type(arr[[True, True]])                                                                                             
Out[44]: text_extensions_for_pandas.array.tensor.TensorArray
```

text_extensions_for_pandas/array/test_tensor.py
BryanCutler: ah ok. It appeared to be the same issue when I first looked, but best to verify all cases.

BryanCutler:
@frreiss I think I worked out all the issues here, please take another look. Thanks!

text_extensions_for_pandas/array/test_tensor.py
frreiss: Any selection using a Boolean mask should return a TensorArray, including cases where the mask contains zero or one `True` values. Pandas follows those semantics for all existing types. For example:
``` python
>>> s = pd.Series([1, 2])
>>> type(s[[True, False]])
pandas.core.series.Series
>>> type(s.values[[True, False]])
numpy.ndarray
```
The same applies to lists of indexes, even of those lists contain exactly zero or one elements:
``` python
>>> type(s[[0]])
pandas.core.series.Series
```

text_extensions_for_pandas/array/test_tensor.py
frreiss: All of these expressions need to return the same type, both for consistency with the way Pandas handles selections with Boolean masks; and to avoid causing headaches for user code that computes a Boolean mask in a data-dependent manner. The code:
``` python
my_df[colname][<expression that returns a Boolean mask>]
```
should always evaluate an object of the same type; otherwise later code will receive unexpected types.

BryanCutler:
Thanks for the careful review @frreiss , I think I have it straightened out now

BryanCutler:
Thanks Fred!

Issue #5, opened 2020-04-03
CoNLL demo, part 3


frreiss:
At the end of part 2 of the demo, we've shown that there are incorrect labels hidden in the CoNLL-2003 validation set, and that you can pinpoint those incorrect labels by data-mining the results of the 16 models the competitors submitted.

Our goal for part 3 of the demo is to pinpoint incorrect labels across the entire data set. The (rough) process to do so will be:
1. Retokenize the entire corpus using a "BERT-compatible" tokenizer, and map the token/entity labels from the original corpus on to the new tokenization.
1. Generate BERT embeddings for every token in the entire corpus in one pass, and store those embeddings in a dataframe column (of type `TensorType`) alongside the tokens and labels.
1. Use the embeddings to *quickly* train multiple models at multiple levels of sophistication (something like: SVMs, random forests, and LSTMs with small and large numbers of hidden states). Split the corpus into 10 parts and perform a 10-fold cross-validation.
1. Repeat the process from part 2 on each fold of the 10-fold cross-validation, comparing the outputs of every model on the validation set for each fold.
1. Analyze the results of the models to pinpoint potential incorrect labels. Inspect those labels manually and build up a list of labels that are actually incorrect.

PR #20, opened 2020-05-28
Next stage of CoNLL scenario

frreiss:
This PR adds several of the planned steps to the CoNLL-2003 scenario (see issue #5):
* Retokenize with a BERT tokenizer,
* Align spans between the BERT and CoNLL tokenizations
* Generates BERT embeddings
* Train a single baseline model over the BERT embeddings
* Train multiple reduced-quality models on the embeddings
* Preliminary examination of the outputs of the models to ensure we still can localize incorrect labels when we generate the multiple models ourselves

Most of the changes are in `CoNLL_3.ipynb`. I've also extended the code in `text_extensions_for_pandas/io/tokenization.py` for aligning spans across tokenizations.


text_extensions_for_pandas/io/tokenization.py
```diff
@@ -70,28 +74,171 @@ def make_bert_tokens(target_text: str, tokenizer) -> pd.DataFrame:
...
+    if len(seq.shape) != 1:
+        raise ValueError(f"Input array must be 1D; got shape {seq.shape}")
+    window_length, pre_padding, post_padding = _compute_padding(
+        len(seq), overlap, non_overlap
+    )
+
+    # First generate the windows as a padded flat arrays.
+    padded_length = len(seq) + pre_padding + post_padding
+    buf = np.zeros(shape=[padded_length], dtype=seq.dtype)
+    buf[pre_padding : pre_padding + len(seq)] = seq
```
BryanCutler, line 137: would `np.concatenate` be better than making a copy of `seq`?

text_extensions_for_pandas/io/test_tokenization.py
```diff
@@ -76,3 +79,28 @@ def test_make_bert_tokens(self):
...
+
+    def test_seq_to_windows(self):
+        for seqlen in range(1, 20):
+            seq = np.arange(1, seqlen)
+            seq_after = windows_to_seq(seq, seq_to_windows(seq, 2, 3)["input_ids"], 2, 3)
+            if np.any(seq != seq_after):
+                raise ValueError("Before: {seq}; After: {seq_after}")
+
+        for seqlen in range(200, 400):
+            # print(f"seqlen: {seqlen}")
```
BryanCutler, line 91: leftover print statements?

text_extensions_for_pandas/io/tokenization.py
frreiss: I don't think there would be any performance benefit to using `np.concatenate` here. As far as I'm aware, `np.concatenate` always allocates a fresh, contiguous buffer and copies the arrays being concatenated into that buffer; see source code at https://github.com/numpy/numpy/blob/50ce0fce70cac779919d97578381a9d762f42594/numpy/core/src/multiarray/multiarraymodule.c#L363

text_extensions_for_pandas/io/test_tokenization.py
frreiss: Good catch! I've removed these statements in dbb7da9

PR #21, opened 2020-05-30
Enable chunked arrays for TensorArray serialization

### LINKED ISSUE #19: ENH: Implement converting large TensorArrays from Arrow
The current Arrow conversion code can convert large `TensorArray` columns to Arrow. Pandas takes care of dividing large values into multiple blocks. However, the conversion code cannot read back large `TensorArray` values that the user has written out this way, because input from multi-block `ArrowRecordBatch`es  is not currently implemented. See [line 256 of `arrow_conversion.py`](https://github.com/CODAIT/text-extensions-for-pandas/blob/ba55ef94e77e768650208c53769b6320decd4739/text_extensions_for_pandas/array/arrow_conversion.py#L256):

``` python
        if extension_array.num_chunks > 1:
            raise ValueError("Only pyarrow.Array with a single chunk is supported")
```

This limitation prevents us from using binary I/O to save precomputed embeddings in part 3 of the CoNLL demo (see #5). Regenerating BERT embeddings for the entire corpus takes about an hour.

BryanCutler:
This allows serialization with Arrow for TensorArray columns with multiple chunks.

Fixes #19

### DIFF
--- a/text_extensions_for_pandas/array/arrow_conversion.py
+++ b/text_extensions_for_pandas/array/arrow_conversion.py
@@ -400,13 +400,23 @@ class ArrowTensorArray(object):
 
     @staticmethod
     def to_numpy(pa_ext_array):
+
+        def make_numpy_array(ext_arr):
+            ext_type = ext_arr.type
+            ext_list_type = ext_arr.storage.type
+            assert pa.types.is_list(ext_list_type)
+            ext_dtype = ext_list_type.value_type.to_pandas_dtype()
+            buf = ext_arr.storage.buffers()[3]
+            return np.ndarray(ext_type.shape, buffer=buf, dtype=ext_dtype)
+
         if isinstance(pa_ext_array, pa.ChunkedArray):
             if pa_ext_array.num_chunks > 1:
-                raise ValueError("Only pyarrow.Column with single array chunk is supported")
-            pa_ext_array = pa_ext_array.chunk(0)
-        ext_type = pa_ext_array.type
-        ext_list_type = pa_ext_array.storage.type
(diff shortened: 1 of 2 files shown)


text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -400,13 +400,20 @@ def iter_gen():
...
-        ext_type = pa_ext_array.type
-        ext_list_type = pa_ext_array.storage.type
-        # assert ext_list_type is pa.list_
-        ext_dtype = ext_list_type.value_type.to_pandas_dtype()
-        buf = pa_ext_array.storage.buffers()[3]
-        return np.ndarray(ext_type.shape, buffer=buf, dtype=ext_dtype)
+                result = np.concatenate([make_numpy_array(chunk)
+                                         for chunk in pa_ext_array.iterchunks()])
+            else:
+                result = make_numpy_array(pa_ext_array.chunk(0))
```
BryanCutler, line 417: Just realized I forgot to handle the case when this is not a chunked array. Not sure when that can come up using with pandas, but might be possible.

BryanCutler:
Thanks @frreiss , I meant to test it out on the large feather file from the notebook, but it looks like there is another issue:
```    
    408             ext_dtype = ext_list_type.value_type.to_pandas_dtype()
    409             buf = ext_arr.storage.buffers()[3]
--> 410             return np.ndarray(ext_type.shape, buffer=buf, dtype=ext_dtype)
    411 
    412         if isinstance(pa_ext_array, pa.ChunkedArray):

TypeError: buffer is too small for requested array
```
I'll look into this as well

PR #26, opened 2020-06-06
Next stage of CoNLL data scenario

frreiss:
This PR contains code needed to implement the fourth part of the CoNLL demo scenario. This part involves setting up a 10-fold cross-validation on the CoNLL-2003 data set, training an ensemble models on the training set of each fold, and evaluating those models on the test set of each fold. The main changes to implement these steps are in the final commit, f572b24, in the notebook `notebooks/CoNLL_4.ipynb`.

The PR also includes refactoring and bug fixes that I made in preparation for implementing `CoNLL_4.ipynb`:

* Fix for broken regression test in `text_extensions_for_pandas.io`: c947860
* Factor out code that is shared across multiple notebooks from the CoNLL scenario into `util.py` and the `text_extensions_for_pandas` package itself: e35782c, 8a79c37
* Fix inconsistencies between the `__equals__` and `__hash__` functions of our extension types for spans. Before these fixes, some of the equality operations were not symmetric, and some of the hash functions would return different hash values for spans that were equal according to `__equals__`: b374720, e9b7525
* Implement the Consolidate operator from spanner algebra, which turned out to be necessary for the post-processing of results from BERT-based named entity models: 644575c
* Add a notebook just for viewing individual CoNLL documents, so you can browse the documents that correspond to outputs in any of the CoNLL demo notebooks without having to recompute the main notebooks:  09f8279
* Fix a bug in `conll_2003_to_dataframes()` that was messing up the whitespace around left parentheses: 8496555
* Fix a bug in `conll_2003_to_dataframes()` that was messing up the whitespace around apostrophes: 5904ce9

I've also added the JupyterLab table of contents extension to the project's default JupyterLab config, in commit dafaf54. This extension makes it much easier to navigate the longer notebooks from the CoNLL demo. You'll need to rerun `env.sh` to pick up that change in your local `pd` environment (or manually run the lines I've added to `env.sh`).


text_extensions_for_pandas/array/char_span.py
```diff
@@ -73,9 +73,19 @@ def __repr__(self) -> str:
...
     def __eq__(self, other):
-        return (self.begin == other.begin
-                and self.end == other.end
-                and self.target_text == other.target_text)
+        if isinstance(other, CharSpan):
+            return (self.begin == other.begin
+                    and self.end == other.end
+                    and self.target_text == other.target_text)
+        elif isinstance(other, CharSpanArray):
+            return other == self
```
BryanCutler, line 81: isn't `self` an instance of `CharSpan`? So this is comparing a scalar to an array?

text_extensions_for_pandas/array/char_span.py
```diff
@@ -73,9 +73,19 @@ def __repr__(self) -> str:
...
+        else:
+            # Different type ==> not equal
+            return False
+
+    def __hash__(self):
+        result = hash((self.target_text, self.begin, self.end))
```
BryanCutler, line 87: would it make sense to cache a hash of the target text and use that whenever computing this hash, and also for equality checking?

text_extensions_for_pandas/array/char_span.py
frreiss: I think comparison of an array and a scalar value is supposed to return a Boolean mask (that is, implicitly broadcast the scalar value to an array the same shape as the array argument) regardless of the order of operations. Are you aware of any documentation to the contrary?

text_extensions_for_pandas/array/char_span.py
frreiss: Yes, adding caching to the hash function would make sense. Created #27 to cover this.

text_extensions_for_pandas/array/char_span.py
BryanCutler: Oh ok, makes sense. It just looked a little strange at first glance

Issue #24, opened 2020-06-02
ENH: Extend CoNLL format I/O to handle files arbitrary extra tags per token


frreiss:
The CoNLL data set we have been using internally contains just the annotations, i.e.:

```
-DOCSTART- O

CRICKET O
- O
LEICESTERSHIRE I-ORG
TAKE O
OVER O
AT O
```

Other copies of this data set available on the internet contain part of speech tags and noun phrase annotations, i.e.:

```
-DOCSTART- -X- -X- O

CRICKET NNP I-NP O
- : O O
LEICESTERSHIRE NNP I-NP I-ORG
TAKE NNP I-NP O
OVER IN I-PP O
AT NNP I-NP O
```
(see https://github.com/patverga/torch-ner-nlp-from-scratch/blob/master/data/conll2003/eng.testa for example)

The API `conll_2003_to_dataframes` in `text_extensions_for_pandas/io/conll.py` should be extended to handle both of these formats. Ideally, that API should be extended to handle an arbitrary mix of single-token and IOB-format columns, so that other data sets in "CoNLL format", such as the ones at https://github.com/juand-r/entity-recognition-datasets, can be read using that API. After these changes, the signature of `conll_2003_to_dataframes()` should look something like this:
```python
def conll_2003_to_dataframes(input_file: str,
                             column_names: List[str],
                             iob_columns: List[bool],
                             space_before_punct: bool = False)\
        -> List[pd.DataFrame]:
    """
     [. . .]
    :param column_names: Names for the metadata columns that come after the 
     token text. These names will be used to generate the names of the dataframe
     that this function returns.
    :param iob_columns: Mask indicating which of the metadata columns after the
     token text should be treated as being in IOB format. If a column is in IOB format,
     the returned dataframe will contain *two* columns, holding IOB2 tags and 
     entity type tags, respectively. For example, an input column "ent" will turn into
     output columns "ent_iob" and "ent_type".
     [. . .]
    """
```

The `conll_2003_output_to_dataframes` function should also be changed so that it passes through the additional data columns that the modified `conll_2003_to_dataframes` function produces.

[...]

PR #30, opened 2020-06-11
Extend CoNLL I/O to support multiple metadata fields

frreiss:
This pull requests contains the changes outlined in #24 to extend our I/O functions for the CoNLL format to support input files that contain more than one metadata field per token.

I have also added regression tests of the new functionality.

I have also reworked the CoNLL-related notebooks so that they use the CoNLL-2003 corpus in its original format, with additional tags associated with each token. Instead of requiring users to obtain their own copy of the corpus, the notebooks now download from a Github repository (not our repo) which hosts a copy of the dataset. This change also removes the mixup of the contents of `eng.testa` and `eng.testb` that was present in our IBM-internal corpus.

While modifying the notebooks to use the downloaded corpus, I also changed the fold names to match the names in the original CoNLL 2003 paper. The fold names are now "train", "development" (or "dev" for short), and "test". The "test" fold is the one that was used to score the original competition.


frreiss:
Pushed remaining changes and removed "WIP" from title.

notebooks/util.py
```diff
@@ -229,7 +265,7 @@ def train_reduced_model(X: np.ndarray, Y: np.ndarray, n_components: int,
         )),
         ("mlogreg", sklearn.linear_model.LogisticRegression(
             multi_class="multinomial",
-            verbose=1,
+            # verbose=1,
```
BryanCutler, line 268: remove?

frreiss: Removed, thanks!

Issue #31, opened 2020-06-18
BUG: Document text with two dollar signs doesn't render properly in Jupyter


frreiss:
Steps to reproduce:
1. Open `CoNLL_View_Doc.ipynb` in this project, which views individual documents from the CoNLL-2003 corpus.
1. Point the viewing code at document **35** of the **dev** fold of the corpus

You'll see something like this:
![display_bug](https://user-images.githubusercontent.com/12436991/85050996-c8ef2200-b14b-11ea-855e-368982c656e7.png)

Note the italic text without spaces. That text is supposed to read "... for $31 a share, or $135 million ...". Instead, JupyterLab is interpreting the dollar signs as the boundaries of a region of embedded LaTeX code.

The `_repr_html_()` method of `CharSpanArray` and `TokenSpanArray` should return a string that renders properly, even when the document contains dollar signs. The current code *does* escape dollar signs (the above snippet turns into `for &#36;31 a share, or &#36;135 million`), but some component of JupyterLab drills through the escaping and treats the `&#36;`'s and everything between them as LaTeX.

A simpler way to reproduce the core problem is to run the following code in a JupyterLab notebook:
``` python
class ReproduceBug: 
    def _repr_html_(self):
        return "for &#36;31 a share, or &#36;135 million"

ReproduceBug()
```

Recommended steps to fix this problem:
1. Figure out what modifications to the above `ReproduceBug` class's `_repr_html_` method will make JupyterLab display the target text with dollar signs
1. Modify the function `util.pretty_print_html()` under `text_extensions_for_pandas` so that the logic in that function for escaping special characters follows whatever procedure worked in the previous step
1. Add a new regression test (or modify an existing regression test) to cover the current problem with displaying dollar signs and ensure that the problem does not recur. Existing regression tests can be found in `text_extensions_for_pandas/test_util.py`

frreiss:
Closed by #32.

PR #32, opened 2020-06-19
Added workaround to avoid double dollar signs not rendering properly

ZachEichen:
Bug fix for issue #31 
Avoids the issue where double dollar signs inside of text are interpreted by the math rendering system built into Jupyter as LaTeX inline math delimiters, causing improper rendering of text. @frreiss

### DIFF
--- a/text_extensions_for_pandas/util.py
+++ b/text_extensions_for_pandas/util.py
@@ -62,7 +62,8 @@ def pretty_print_html(column: Union["CharSpanArray", "TokenSpanArray"]) -> str:
             text_pieces.append("&#39;")
         elif text[i] == "$":
             # Dollar sign messes up Jupyter's JavaScript UI.
-            text_pieces.append("&#36;")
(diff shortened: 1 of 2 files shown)


text_extensions_for_pandas/util.py
```diff
@@ -62,7 +62,8 @@ def pretty_print_html(column: Union["CharSpanArray", "TokenSpanArray"]) -> str:
             text_pieces.append("&#39;")
         elif text[i] == "$":
             # Dollar sign messes up Jupyter's JavaScript UI.
-            text_pieces.append("&#36;")
+            # Use backslash to avoid
```
frreiss, line 65: I think you meant to say "Embed a `<span>` element to prevent Jupyter from interpreting the dollar sign as the beginning of some embedded LaTex" in this comment; see issue #31 " here

frreiss:
LGTM. Merging.

Issue #28, opened 2020-06-10
BUG: Feather-related regression tests failing at head of master branch


frreiss:
When I run `python -m unittest discover` at the head of the master branch, 5 regression tests related to Feather I/O fail. Output follows:

```
======================================================================
ERROR: test_feather (text_extensions_for_pandas.array.test_char_span.CharSpanArrayIOTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/freiss/pd/tep-conll-4/text_extensions_for_pandas/array/test_char_span.py", line 396, in test_feather
    pd.testing.assert_frame_equal(df, df_read)
  File "/Users/freiss/opt/miniconda3/envs/pd/lib/python3.7/site-packages/pandas/_testing.py", line 1382, in assert_frame_equal
    obj=f'{obj}.iloc[:, {i}] (column name="{col}")',
  File "/Users/freiss/opt/miniconda3/envs/pd/lib/python3.7/site-packages/pandas/_testing.py", line 1184, in assert_series_equal
    assert_extension_array_equal(left.array, right.array)
  File "/Users/freiss/opt/miniconda3/envs/pd/lib/python3.7/site-packages/pandas/_testing.py", line 1034, in assert_extension_array_equal
    right_valid = np.asarray(right[~right_na].astype(object))
  File "/Users/freiss/opt/miniconda3/envs/pd/lib/python3.7/site-packages/pandas/core/arrays/base.py", line 443, in astype
    return np.array(self, dtype=dtype, copy=copy)
  File "/Users/freiss/opt/miniconda3/envs/pd/lib/python3.7/site-packages/pandas/core/arrays/base.py", line 352, in __iter__
    yield self[i]
  File "/Users/freiss/pd/tep-conll-4/text_extensions_for_pandas/array/char_span.py", line 297, in __getitem__
    int(self._ends[item]))
  File "/Users/freiss/pd/tep-conll-4/text_extensions_for_pandas/array/char_span.py", line 65, in __init__
    raise ValueError(f"end must be less than length of target string "
ValueError: end must be less than length of target string (32 > 15

======================================================================
ERROR: test_feather (text_extensions_for_pandas.array.test_tensor.TensorArrayIOTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/freiss/pd/tep-conll-4/text_extensions_for_pandas/array/test_tensor.py", line 346, in test_feather
    df_read = pd.read_feather(filename)
[...]

PR #35, opened 2020-06-22
Pin pyarrow version at 0.17.0 as a workaround for an Arrow bug

frreiss:
This pull request temporarily pins the version of our `pyarrow` dependency at 0.17.0 as a workaround for the problems reported in #28 . The changes to `requirements.txt` and the `env.sh` script are in commit 613b231

I also needed to change the SpaCy-related regression tests a bit to get them to pass with the latest version of the SpaCy English small language model. I've included those changes as a second commit, 4ffe713.

Once pyarrow 1.0.0 is released, we will want to update `requirements.txt` to use that version, which has a fix for the Arrow bug that is the root cause of #28 .


ZachEichen:
Looks good to me.

PR #37, opened 2020-06-23
Add cell that displays tokens

frreiss:
This PR adds an additional cell with token information to the notebook for displaying documents.


frreiss:
Merging directly, since this is a very small change.

PR #42, opened 2020-07-01
Update BERT tokenizer integration to reflect API changes in Transformers library

frreiss:
The latest version of the [Transformers library](https://huggingface.co/transformers/) moved the `PreTrainedTokenizerFast` class to a new location within the package. This change breaks our integration with that project's tokenizer.

This PR updates the `import` statement to use the new location of `PreTrainedTokenizerFast`. I have also modified `dev_reqs.txt` to specify version 3.0.0 or greater of `transformers`.

@BryanCutler @xuhdev @ZachEichen you will want to rerun `env.sh` or manually upgrade your local copy of the `transformers` Python library after this PR is merged.

PR #43, opened 2020-07-02
Make table of span offsets optional in _repr_html

frreiss:
This PR adds a flag to turn off the table of detailed spans in the HTML (i.e. Jupyter notebook) representation of `TokenSpanArray` and `CharSpanArray`.  Removing the table is useful when the user is only interested in seeing the marked-up target text. The flag is called `repr_html_show_offsets` and is a separate property for each object.

The default value of `True` gives the original `_repr_html()` behavior:
repr_html_before

Setting the flag to `False` removes the table of span offsets:
repr_html_after

I have also updated the regression tests accordingly and eliminated some code duplication between `CharSpanArray` and `TokenSpanArray`.


text_extensions_for_pandas/array/token_span.py
```diff
@@ -764,12 +764,6 @@ def as_frame(self) -> pd.DataFrame:
     ##########################################
     # Keep private and protected methods here.
 
-    def _repr_html_(self) -> str:
```
ZachEichen, line 766: why is this being removed?

ZachEichen: other than this it looks good to me!

frreiss: That function overrides a function by the same name, with the same function body, in the parent class `CharSpanArray`. I discovered the duplicate functions when I modified `CharSpanArray._repr_html()`. I think the duplicate functions are a relic of the time before `TokenSpanArray` was a subclass of `CharSpanArray`.

frreiss:
Build passes now after merging #42. Merging this PR.

PR #45, opened 2020-07-08
Do not override sys.path in notebooks

xuhdev:
`sys.path[0]` can be an empty string (meaning the script's dir), and `import util` is failing if it is the case.


frreiss:
LGTM. Build failures appear to be due to SpaCy messing with their models again and unrelated to the changes.

PR #46, opened 2020-07-08
Add copy constructor and test of sum aggregates

frreiss:
This PR adds a copy constructor to the `TensorArray` class. I added tests of the copy constructor and also threw in some tests of `sum` aggregations.

### DIFF
--- a/text_extensions_for_pandas/array/test_tensor.py
+++ b/text_extensions_for_pandas/array/test_tensor.py
@@ -55,6 +55,10 @@ class TestTensor(unittest.TestCase):
         with self.assertRaises(TypeError):
             TensorArray(2112)
 
+        # Copy constructor
+        s_copy = TensorArray(s)
+        self.assertEqual(len(s), len(s_copy))
+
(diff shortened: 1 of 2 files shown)


frreiss:
Test failures on this PR are due to #47

text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -55,6 +55,10 @@ def test_create(self):
         with self.assertRaises(TypeError):
             TensorArray(2112)
 
+        # Copy constructor
+        s_copy = TensorArray(s)
```
BryanCutler, line 59: Is there a need to use like a copy constructor? I thinks it's more common to explicitly copy like `TensorArray(s).copy()`

frreiss: Makes sense. Updating this PR accordingly.

frreiss:
Thanks for the review, @BryanCutler ! Pushed some changes to address review comments.

PR #49, opened 2020-07-08
Add scikit-learn and torch to notebooks/jupyter_reqs.txt

xuhdev:
torch and scikit-learn are used in notebooks/util.py


frreiss:
LGTM. CI failures due to continued problems with pustream SpaCy models changing. Merging.

PR #50, opened 2020-07-09
Minor changes

frreiss:
Minor update to markdown, plus change the Python version in an `ipynb` file to the latest build of Python 3.7 to prevent diffs from happening in the future.


frreiss:
This is not a big enough change to warrant a review. Merging.

Issue #44, opened 2020-07-02
Extend Watson NLU notebook to also use SpaCy


frreiss:
The demo notebook `notebooks/WatsonNLP_Demo.ipynb` in #25 puts a document through Watson Natural Language Understanding and converts the results from Watson NLU into a collection of dataframes.

Currently the last step of this notebook involves taking the outputs of relationship extraction and connecting those outputs to sentence information from another part of the Watson NLU output.

We should extend this notebook with some additional steps to illustrate both our integration with SpaCy and how Text Extensions for Pandas makes it easier to use multiple NLP frameworks in the same application. Here's a rough outline of the proposed additional steps:
1. Pass the document text (should be available either in a local variable or as the target text of a CharSpanArray object) to SpaCy's English language model, and convert the resulting graph of Python objects to a dataframe (See example code in `Person.ipynb`).
2. Match the entities involved in relationships that Watson NLU returns with the corresponding SpaCy tokens from the SpaCy language model output.
3. Now we have a connection from entities to sets of 1 or more tokens. Trace the parse tree links in the SpaCy output (encoded in the `head` column of the dataframe) to identify the least common ancestor of both sides of the relationship. Then pull out the notes of the parse tree that are below this common ancestor. The output of this operation should of course be a dataframe (containing all the nodes in the subtree of the parse tree). The tracing itself should if possible be implemented with a Gremlin query or two.
4. Use our integration with DisplaCy to show the relevant subparts of the parse tree. The entry point to use is `tp.render_parse_tree()`. See the example usage at the bottom of `Person.ipynb` for more info.

frreiss:
Covered in `Integrate_NLP_Libraries.ipynb`.

PR #52, opened 2020-07-10
Extend Watson NLU notebook to also use SpaCy

ZachEichen:
Addresses issue #44, by combining use of SpaCy with Watson NLU, and look at parse trees of relations found by Watson NLU engine.

Also fixes a bug found in the recurse.py in the gremlin engine, which caused emit recursions to not repeat.


frreiss:
LGTM, merging.

PR #55, opened 2020-07-17
Update code for plotting F1 score

frreiss:
This code modifies the F1 score plotting code in two of the CoNLL related notebooks to generate plots that are easier to read and more consistent with each other.

Apologies for the large diff; I had to recompute the notebooks.


frreiss:
This is a very minor update, so not bothering with a review. Merging.

PR #58, opened 2020-07-22
Fix tests again

frreiss:
SpaCy shipped a new version of their library, with new language models that produce substantially different results even on simple text. This PR updates our regression tests to account for the latest SpaCy model changes.

I've also pinned the version of SpaCy in `dev_reqs.txt` to *exactly* 2.3.2 in the hopes that doing so will prevent this sort of thing from happening in the future.


frreiss:
FYI @ZachEichen and @BryanCutler you will want to recreate your local Anaconda environments after these changes are merged.

PR #67, opened 2020-07-29
Implement remaining hooks for TensorArray

BryanCutler:
Add remaining hooks for TensorArray. Calling `pandas.Series.factorize` still requires `_values_for_factorize` to be implemented.

### DIFF
--- a/text_extensions_for_pandas/array/tensor.py
+++ b/text_extensions_for_pandas/array/tensor.py
@@ -107,7 +107,7 @@ class TensorArray(pd.api.extensions.ExtensionArray, TensorOpsMixin):
                  make_contiguous: bool = True):
         """
         :param values: A `numpy.ndarray` or sequence of `numpy.ndarray`s of equal shape.
-        :param make_contiguous: force values to be contiguous in memory
+        :param make_contiguous: force values to be contiguous in memory (C order)
         """
         if isinstance(values, np.ndarray):
             self._tensor = values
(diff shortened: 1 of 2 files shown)


text_extensions_for_pandas/array/tensor.py
```diff
@@ -224,6 +263,14 @@ def __repr__(self):
     def __str__(self):
         return self._tensor.__str__()
 
+    def _values_for_factorize(self) -> Tuple[np.ndarray, Any]:
+        """
+        See docstring in `ExtensionArray` class in `pandas/core/arrays/base.py`
+        for information about this method.
+        """
+        # TODO return self._tensor, np.nan
+        raise NotImplementedError
```
BryanCutler, line 272: not sure how useful this will be, but I think there should be something here for completness

text_extensions_for_pandas/array/tensor.py
```diff
@@ -214,6 +247,12 @@ def __setitem__(self, key: Union[int, np.ndarray], value: Any) -> None:
             raise NotImplementedError(f"__setitem__ with key type '{type(key)}' "
                                       f"not implemented")
 
+    def __eq__(self, other):
+        """
+        Return self==value.
+        """
+        return self == other
```
frreiss, line 254: Won't this lead to infinite recursion?

text_extensions_for_pandas/array/tensor.py
frreiss: Yeah, you'll probably find that part of the Pandas suite for extension types fails if this method isn't implemented.

text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -260,6 +260,12 @@ def test_sum(self):
         sum_some = df["s"][[True, False, True]].sum()
         npt.assert_array_equal(sum_some.to_numpy(), [6, 8])
 
+    def test_factorize(self):
+        x = np.array([[1, 2], [3, 4], [5, 6], [3, 4]])
+        s = TensorArray(x)
+        with self.assertRaises(NotImplementedError):
+            indices, values = s.factorize()
+
```
frreiss, line 268: Can you add a test of `__equals__` here too?

text_extensions_for_pandas/array/tensor.py
BryanCutler: I forgot the mixin already defines this and others. It actually overrides this so this particular method is never called. Taking it out.

text_extensions_for_pandas/array/test_tensor.py
BryanCutler: I have a test that does all the binary operators, so it was being tested there

BryanCutler:
Thanks @frreiss , I'll go ahead and merge if tests pass and fix the remaining issues when we get the pandas test suite running.

Issue #18, opened 2020-05-20
ENH: Create Dataframe wrapper for Watson table extraction API


frreiss:
Create a wrapper for the Watson Compare and Comply table extraction APIs as part of the "Watson Natural Language Understanding JSON output" line item from our version 0.1 roadmap (#1)

API docs are here: https://cloud.ibm.com/apidocs/compare-comply#extract-a-document-s-tables

The API doc page includes an example of the JSON output of the API (`element_classification`). The dataframe version of the API output should include all information present in the JSON.

frreiss:
This wrapper work can proceed in two phases:
* First directly wrap the output schema of the REST API. That is, the `table_headers` section of the REST API's return value turns directly into a dataframe called "table_headers"
* Then imlement heuristics that will reconstruct a Pandas dataframe that looks like the original table (as opposed to the "shredded" representation that the API returns)

frreiss:
Example documents to use as a target for demonstrating this functionality: https://www.who.int/emergencies/diseases/novel-coronavirus-2019/situation-reports/

frreiss:
There is general purpose code in https://github.com/CODAIT/text-extensions-for-pandas/blob/master/text_extensions_for_pandas/io/watson.py that you should be able to use for the first phase of this work item; directly wrapping the output schema of the REST API in a set of Pandas dataframes.

frreiss:
#68 closes out this issue.

PR #68, opened 2020-07-31
ADD: Dataframe wrapper for Watson table extraction API

ZachEichen:
created a module, `watson_tables` in the `io` module that reads output from the Watson Compare & Comply table extraction and understanding.  (addresses issue #18 )

This is designed to be an easy to use package that works in two main steps: 

- First use the `watson_tables_parse_response()` method to directly wrap the contents of the API response. This is more or less a 1:1 representation, but in pandas dataframe format, and is composed of 3 sub-elements; a DataFrame describing column headers, a DataFrame describing row headers, and a DataFrame describing the body cells of the table. 

- Then the `make_table()`  method can be called on the response from the first portion, where it applies a series of heuristics to convert the plainly wrapped and still deconstructed output into a fully featured pandas DataFrame that can be used to manipulate and analyze the data. 

The second step can further be broken down into two user-accessable phases: 
 - creation of the "exploded" or value attribute mapping dataframe. This maps the value of each cell with its attributes: i.e. its row and column information and applicable column headers 
 - It ingests that "exploded" format, and uses it to construct a version of the original table in pandas DataFrame format, where it can be used to manipulate data

The files that have been added are as follows: 
1.  `watson_tables.py` - created module, and added functionality
1. `test_watson_tables.py` created test cases for `watson_tables` module
1.  ` archive.json` (one in resources and one in test data) json archive of Watson table extraction API responses for demo and testing purposes, respectively 
1.  pdf files in resources/tables: pdf files containing tables for demo.


frreiss:
This looks good overall. Almost ready to merge.

One thing to change: I think it would be best for us not to include the original PDFs source for the WHO data. The license on those files (https://www.who.int/about/who-we-are/publishing-policies/copyright) may not be compatible with the license on the `text-extensions-for-pandas` project.

Can you please remove `resources/tables/WHO-covid-overview-chart.pdf` and `resources/tables/who-covid-report-full.pdf` and instead add a markdown file to that directory with instructions for downloading those files directly from the World Health Organization web site? The main download page is at https://www.who.int/emergencies/diseases/novel-coronavirus-2019/situation-reports

I think that keeping the processed output from the WHO table inside `archive.json` should be ok.

ZachEichen:
The who files are removed in the latest commit

frreiss:
Thanks @ZachEichen ! Merging.

PR #69, opened 2020-08-03
Add Pandas extension array test suites for TensorArray

BryanCutler:


requirements.txt
```diff
@@ -5,4 +5,4 @@ regex
 fastparquet
 # TODO: The following dependency should go away when we switch to Python 3.8.
 memoized-property
-
+pytest
```
BryanCutler, line 8: Pandas uses fixtures which requires pytest

BryanCutler:
@frreiss , I have added all the test suites for TensorArraybut there got to be  too many errors, so I'll have to work through them one by one and see what we can support. The other extension types should be easier I believe. Adding these tests does require pytest because of the use of fixtures, so wanted to make sure you're ok with that.

frreiss:
@BryanCutler sure, adding `pytest` is fine.

text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -299,11 +313,14 @@ def test_sum(self):
...
-                b    [[1, 1], [1, 1]]
-                c    [[3, 3], [3, 3]]"""
+                             value
+                key               
+                a   [[2 2]
+                 [2 2]]
+                b   [[1 1]
+                 [1 1]]
+                c   [[3 3]
+                 [3 3]]"""
```
BryanCutler, line 323: The change in `_from_sequence` caused this. This looks worse, but `_from_sequence` was wrong before so we can figure out how to format this better later.

BryanCutler:
@frreiss are you ok with merging this as is and I can start working on adding tests for the other extension arrays?

text_extensions_for_pandas/array/test_tensor.py
frreiss: ok

frreiss:
Sure, this looks OK to merge.

BryanCutler:
Thanks @frreiss !

PR #71, opened 2020-08-05
Add Pandas base extension tests for CharSpan

BryanCutler:


text_extensions_for_pandas/array/char_span.py
```diff
@@ -69,8 +72,11 @@ def __init__(self, text: str, begin: int, end: int):
         self._end = end
 
     def __repr__(self) -> str:
-        return "[{}, {}): '{}'".format(self.begin, self.end,
-                                       textwrap.shorten(self.covered_text, 80))
+        if self.begin == CharSpan.NULL_OFFSET_VALUE:
+            return "None"
+        else:
+            return "[{}, {}): '{}'".format(self.begin, self.end,
+                                           textwrap.shorten(self.covered_text, 80))
```
BryanCutler, line 79: @frreiss  this was causing a problem when `begin` as CharSpan.NULL_OFFSET_VALUE which makes `self.covered_text` None and causes textwrap.shorten to crash. Let me know if you think it should print something other than "None" for that case.

frreiss: I'd say it's best here to synchronize with how Pandas represents missing values. The [docs](https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html) seem to point to "NA" being the standard representation for missing values in new code.

BryanCutler: Yeah, I agree. "NA" will be better.

BryanCutler:
@frreiss I think I enabled all the Pandas tests that we could reasonably support for CharSpanArray. Some are skipped because they are not applicable, like arithmetic ops, and some could be enabled but are not too vital and require more effort, such as parsing tests that require `_from_sequence_of_strings`. If these changes look good to you, I'll so the same for TokenSpanArray, which should be almost identical.

BryanCutler:
Not sure what's going on with the travis build error. Some kind of mixup in pandas, will have to look at it later.

frreiss:
@BryanCutler those build errors were due to Pandas changing some of the internal APIs we use. The PR of mine that you just merged (thanks, BTW!) temporarily pinned Pandas at <= 1.0.5. I'm going to merge the current pull request now. 

After the merge goes through, would you mind taking a look at the tests that break when you upgrade Pandas to version 1.1.0?

BryanCutler:
Thanks @frreiss , sure I will take a look with that version of Pandas

PR #72, opened 2020-08-05
Refactoring: Pull code from notebooks/util.py into main package

frreiss:
This PR pulls several of the functions that had been in `notebooks/util.py` up into the main package. I've added tests for the new functionality and added some new dependencies to our development dependencies to ensure the tests will run.

I also temporarily disabled the warning messages that the tests in `test_watson.py` were putting out. We should probably figure out what is up with those later.


text_extensions_for_pandas/io/test_tokenization.py
```diff
@@ -17,15 +17,33 @@
...
+    conll_2003_to_dataframes, make_iob_tag_categories
+)
+from text_extensions_for_pandas.array import (
+    TensorArray
 )
 
 
 class TestTokenize(unittest.TestCase):
 
+    def __init__(self, *args, **kwargs):
```
BryanCutler, line 39: maybe better to put this in `setupClass(cls)`?

BryanCutler:
I'll take a look at the watson warning and see what's going on

text_extensions_for_pandas/io/test_tokenization.py
frreiss: Very good idea. I will make that change. I have some additional changes to that file already in my local copy, so I'll merge the current set of changes and include the `setupClass()` method in the next batch.

PR #74, opened 2020-08-05
Add better message for Watson sentence mismatch warning

BryanCutler:


BryanCutler:
@frreiss it is giving a warning when trying to replace the sentence col in the watson relations dataframe with a span. It's because the argument locations match the sentence span from the syntax dataframe, but sentence text doesn't match the span text, although they look identical. For Example:
```
Span Text: 
When Arthur and Bedevere reach the bridge's end, they cannot find Lancelot, unaware he has been arrested by the police investigating the historian's death.

Sentence: 
When Arthur and Bedevere reach the bridge's end, they cannot find Lancelot, unaware he has been arrested by the police investigating the historian's death.
``` 

The comparison is done with 
```
sentence_span.covered_text == sentence_col[i]
```
here https://github.com/CODAIT/text-extensions-for-pandas/pull/74/files#diff-ca4207b824350773b4b9bd346a68a3d9R230

I also can't reproduce it locally, any ideas?

frreiss:
Most likely the two strings are different, but they look the same when you print them to the screen.

Some things to try:
* Put quotes around the things you print. There may be leading/trailing whitespace.
* Dump the strings character-by-character as hexadecimal
* Write a comparison function that also outputs an explanation for the difference

As for not being able to reproduce the problem locally, what platforms does this problem occur on?

BryanCutler:
ok figured it out, the watson sentence column is an Arrow array and the scalar value is a pyarrow.StringScalar. After converting that to a python string the warning goes away. Not sure why it's fine on my machine because it's the same pyarrow version, but it's the right thing to do to convert it to a python string anyway. So I think this should be good to merge if it looks ok with you @frreiss ?

BryanCutler:
I take that back, I did have a different pyarrow version in my watson test env and that's why it never came up. I found some related JIRAs https://issues.apache.org/jira/browse/ARROW-9445, and sent a message to the dev list to make sure it wasn't a regression.

PR #75, opened 2020-08-06
ADD: additional functionality to table wrapper, and improved demo notebooks 

ZachEichen:
Added functionality to table wrapper, including auto detection of numeric rows and columns,  and logic to maintain row / column order from the original table, as opposed to reverting to alphabetic order after pivot. 

Also cleaned up and improved the presentability of demo notebooks.


notebooks/Tables.ipynb
```diff
@@ -30,6 +30,7 @@
     "from ibm_watson import ApiException\n",
```
frreiss, line 30: <p>Averaging together different items with different units doesn't really make sense. I'd recommend doing max of the two tax rates to show aggregation along an axis.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/75/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

PR #77, opened 2020-08-13
Update code and tests for pandas >= 1.1.0

BryanCutler:


BryanCutler:
@frreiss I skipped most of the new tests added for now, some will need to be fixed later. Tests passing now and I confirmed CI installed Pandas 1.1.0 https://travis-ci.com/github/CODAIT/text-extensions-for-pandas/builds/179617623#L574

requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy>=1.17
-pandas>=1.0.3,<1.1.0
+pandas>=1.0.3
```
frreiss, line 2: Should this be `>=1.1.0`? Tests currently pass on Pandas 1.0.5, but a lot of them are disabled...

BryanCutler: I was just about to test with lower versions. I think it would be better to have tests with at least 2 version of pandas, how about 1.0.5 and 1.1.0? Shall I setup travis to do this?

PR #79, opened 2020-08-14
Change CI to test 2 versions of Pandas

BryanCutler:


BryanCutler:
@frreiss this ends up downgrading Pandas from 1.1.0 to 1.0.5 in one CI job, but that's not too time consuming and the easiest way to do this without fixing the version in `config/dev_env.yml`

frreiss:
LGTM. Merging.

PR #81, opened 2020-08-18
Add Pandas base extension tests for TokenSpan

BryanCutler:


text_extensions_for_pandas/array/token_span.py
```diff
@@ -271,6 +274,8 @@ def __init__(
         :param begin_tokens: Array of begin offsets measured in tokens
         :param end_tokens: Array of end offsets measured in tokens
         """
+        if not isinstance(tokens, CharSpanArray):
+            raise TypeError(f"Expected CharSpanArray as tokens but got {type(tokens)}")
```
BryanCutler, line 278: This was the only new issue compared with the CharSpan tests. For `test_construct_empty_dataframe`, it will call `_from_sequence` with an empty list that leads to `tokens` here being `NoneType`. Something similar happens for CharSpanArray, but it passes as-is, so it wasn't noticed. Not sure if creating empty arrays is something we should support for these, what are your thoughts @frreiss ?

frreiss: Yes, we absolutely should support zero-length `TokenSpanArray`s. And if Pandas' internal functions can pass `None` to the constructors for extension types, then our extension types need to handle that type of input. I recommend creating a dummy CharSpanArray with a target text set to an empty string for this case.

PR #83, opened 2020-08-19
Add unit tests for spanner extract and project

### LINKED ISSUE #63: Flesh out regression tests of `spanner` module
We are currently missing regression tests for the files: `spanner/extract.py` and `spanner/project.py`. Add basic tests for these two files. It is not necessary at this point to test everything.

BryanCutler:
This adds basic unit tests for the spanner extract and project modules.

Fixes #63

### DIFF
--- a/text_extensions_for_pandas/spanner/test_extract.py
+++ b/text_extensions_for_pandas/spanner/test_extract.py
@@ -1,0 +1,120 @@
+#
+#  Copyright (c) 2020 IBM Corp.
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+import re
+import textwrap
+
+from text_extensions_for_pandas.spanner.extract import *
+from text_extensions_for_pandas.io.spacy import make_tokens
+from text_extensions_for_pandas.io.systemt import load_dict
+from text_extensions_for_pandas.util import TestBase
+
+import spacy
+
+
+class ExtractTest(TestBase):
+
+    @staticmethod
+    def _make_tokenizer():
+        from spacy.lang.en import English
+        nlp = English()
+        return nlp.Defaults.create_tokenizer(nlp)
+
+    @staticmethod
+    def _read_file(file_name):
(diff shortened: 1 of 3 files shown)


text_extensions_for_pandas/spanner/test_extract.py
```diff
@@ -0,0 +1,120 @@
...
+    @staticmethod
+    def _read_file(file_name):
+        with open(file_name, "r") as f:
+            lines = [
+                line.strip() for line in f.readlines() if len(line) > 0 and line[0] != "#"
+            ]
+        return " ".join(lines)
+
+    def test_extract_dict(self):
+        file_name = "test_data/io/test_systemt/test.dict"
```
frreiss, line 44: I think the location of this file is an anachronism. Would you mind moving it to `test_data/spanner`?

frreiss: Also, I'd recommend that you remove the last three lines of the current file and replace "file_text" below with a string that exercises the major cases of dictionary extraction:
* Match at beginning of string, end of string, or in the middle of the string
* One-token match and multi-token match
* Non-match that shares the first token (and only the first token) with a two-token dictionary entry
* Two overlapping matches

text_extensions_for_pandas/spanner/test_extract.py
```diff
@@ -0,0 +1,120 @@
...
+                1                  [17, 22): 'Entry'
+                5  [23, 44): 'Help me! I am trapped'
+                4    [45, 64): 'In a Haiku factory!'
+                3    [65, 84): 'Save me before they'"""
+            )
+        )
+
+    def test_extract_regex_tok(self):
+        file_name = "test_data/io/test_systemt/test.dict"
+        file_text = self._read_file(file_name)
```
frreiss, line 72: As with the dictionary test, it would be useful to have a target string that contains the main types of regex match -- matches at the beginning, middle, or end of the string; partial matches; substrings that would be matches except they don't start or end on a token boundary.

text_extensions_for_pandas/spanner/test_extract.py
frreiss: You'll also want to exercise case-insensitivity of the dictionary matching.

frreiss:
@BryanCutler should we merge the changes from this branch now, or are you still making improvements?

BryanCutler:
@frreiss  I was going to bring this up, but ran out of time today. I'm not entirely clear on how dictionary extraction is supposed to work, which is why the test is pretty simple. I'll take another look and make this better before merging.

text_extensions_for_pandas/spanner/test_extract.py
BryanCutler: > I think the location of this file is an anachronism. Would you mind moving it to test_data/spanner?

This data file is still being used in `test_systemt` also, you mean make a copy of it for spanner and replace the Haiku with something else?

PR #94, opened 2020-08-20
Fix test for astype with string dtype

BryanCutler:


BryanCutler:
This is how it was done in `BooleanArray`

text_extensions_for_pandas/array/char_span.py
```diff
@@ -314,9 +314,13 @@ def astype(self, dtype, copy=True):
         See docstring in `ExtensionArray` class in `pandas/core/arrays/base.py`
         for information about this method.
         """
+        from pandas.core.arrays.string_ import StringDtype
```
frreiss, line 317: Is this different from the `StringDtype` symbol that's exported at the top-level scope of the `pandas` package? See https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.StringDtype.html#pandas.StringDtype

PR #98, opened 2020-08-27
Fix up Watson NLU notebook

frreiss:
This pull request includes a rework of the Watson NLU use case notebook. I have renamed the notebook from `Watson_NLU_Demo.ipynb` to `Analyze_Text.ipynb` and added explanatory Markdown sections, plus some additional examples of manipulating span information, to the original example. I have broken out the SpaCy-related parts at the end of the original notebook into a second notebook, `Combine_Libraries.ipynb`. I replaced the Gremlin examples there with examples that use Pandas operations over spans instead.

I also fixed up a few things along the way:
* Extended the implementation of `__add__()` so that it covers all combinations of span types
* Moved the dependency on `pytest` from `requirements.txt` to the `dev_reqs.txt`
* Added an additional argument to `render_parse_tree()` and `token_features_to_tree()` for specifying the column with head node IDs.


notebooks/Analyze_Text.ipynb
```diff
@@ -0,0 +1,2895 @@
+{
```
BryanCutler, line 1: <p>Might be good to make it clear that the first path is raw data + python and only to show how messy it is compared to the neat Pandas work following. It's a lot for the user to look at and they might not know we are going to show a much better version.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/98/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

BryanCutler, line 1: <p>Maybe wrap the char_span repr in quotes to be easier to pick out?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/98/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

BryanCutler, line 1: <p>It would be good to have a sentence describe what kind of data is in each of the following DataFrames, I could do that as a followup.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/98/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='58'/>

notebooks/Combine_Libraries.ipynb
```diff
@@ -0,0 +1,3663 @@
+{
```
BryanCutler, line 1: <p>Maybe change the title to something like "NLP_Library_Integration"? Combine_Libraries is maybe a little hard to tell what it's referring to.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/98/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

BryanCutler, line 1: <p>did you mean to delete "containing" and end the sentence?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/98/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='46'/>

text_extensions_for_pandas/array/char_span.py
```diff
@@ -634,6 +637,11 @@ def __ge__(self, other):
         # TODO: Figure out what the semantics of this operation should be.
         raise NotImplementedError()
 
+    def __add__(self, other):
+        raise NotImplementedError(
+            "This method should be overwritten by code in span_mixins.py"
```
BryanCutler, line 642: I think it's just `span.py`?

text_extensions_for_pandas/array/span.py
```diff
@@ -0,0 +1,106 @@
...
+                             np.minimum(span1.begin, span2.begin),
+                             np.maximum(span1.end, span2.end))
+    else:
+        raise TypeError(f"Unexpected combination of span types for add operation: "
+                        f"{type(span1)} and {type(span2)}")
+
+
+# Add polymorphic methods to the span classes
+CharSpan.__add__ = lambda self, other: add_spans(self, other)
+CharSpanArray.__add__ = lambda self, other: add_spans(self, other)
```
BryanCutler, line 105: Why not make it a true mixin class that could a subclass of CharSpan and CharSpanArray? This kind of method assignment isn't great to do imho. It's hard to follow and might make it difficult to debug.

notebooks/Analyze_Text.ipynb
frreiss: <p>Good point. I've expanded on the introduction here with a description of the two implementations we're about to do and some warnings that the non-Pandas version of the syntax analysis output is a bit daunting.</p>

frreiss: <p>Good idea. Done.</p>

frreiss: <p>Good point. I've created issue #101 to cover creating this additional text.</p>

notebooks/Combine_Libraries.ipynb
frreiss: <p>Great idea! I think I'll tweak it a bit to "Integrate_NLP_Libraries" so that it's consistent with the format of the other notebook titles ("&lt;action verb&gt;_&lt;target of action&gt;").</p>

frreiss: <p>Thanks for catching that! Fixed now.</p>

text_extensions_for_pandas/array/char_span.py
frreiss: Fixed this as a side-effect of other fixes.

text_extensions_for_pandas/array/span.py
frreiss: After some thought, I've replaced this code with a version that avoids overwriting class methods; see 85ec3c5 . 

I couldn't figure out a way to do a proper mixin without introducing circular file dependencies. I'm open to suggestions.

BryanCutler: Could you just define the mixin in the `char_span.py` file? Unless that is making the file too big, otherwise the current way looks fine to me.

frreiss: I would love to, but I couldn't figure out a way to define the mixin in `char_span.py` without pulling the `TokenSpan*` classes into that file :-\ Let me know if you figure out a way.

PR #99, opened 2020-08-27
Fixes to pass addtional Pandas suite of tests for TensorArray

BryanCutler:


BryanCutler:
There seem to be quite a lot of differences between Pandas versions here, I'll have to take a closer look

text_extensions_for_pandas/array/tensor.py
```diff
@@ -103,7 +105,7 @@ class TensorArray(pd.api.extensions.ExtensionArray, TensorOpsMixin):
     Each tensor must have the same shape.
     """
 
-    def __init__(self, values: Union[np.ndarray, Sequence[np.ndarray]],
+    def __init__(self, values: Union[np.ndarray, Sequence[np.ndarray], Any],
```
BryanCutler, line 108: I couldn't figure out a way to make a type hint that would satisfy the condition `np.isscalar` below, so used `Any` here for now. Any ideas @frreiss ?  This was needed for a Pandas 1.0.x test, and it does make sense that one might construct an array from a single value.

BryanCutler:
This is passing now for both Pandas versions, but had to skip some for Pandas 1.0.x. These might be fixed if we introduce a wrapper for a scalar value of the array, so I don't think we need to bump the minimum Pandas version yet until we can see if that will fix the issue.

text_extensions_for_pandas/array/tensor.py
frreiss: I think `Any` is the best solution here. If you really want to, you can build a "object for which `np.isscalar()` returns true" type using the facilities in [the `typing` package](https://docs.python.org/3/library/typing.html). But that option strikes me as overkill.

BryanCutler:
I'll merge this now and circle back to enabling the remain work later.

PR #100, opened 2020-08-27
Move pytest requirement to dev_reqs.txt

BryanCutler:


BryanCutler:
Just noticed this is done in #98

PR #103, opened 2020-08-31
Refactor span classes to lower users' learning curve

frreiss:
This PR makes some changes to our class names to simplify the learning process for new users. The motivation behind this change is the following principle: *A user who only needs the functionality of character-based spans should not need to understand token-based spans.*

The main changes I've implemented are:
* We now refer to character-based spans as just "spans". The class that used to be called `CharSpan` is now called `Span`; `CharSpanArray` is now called `SpanArray`; and so on.
* Our dtypes now have names that end in "Dtype", for consistency with how Pandas names its data type objects.
* Instead of returning two columns of per-token spans ("token_span" and "char_span"), all the syntax analysis input functions (SpaCy, CoNLL, and Watson NLU) now return just a column "span" of dtype `SpanDtype`
* I've updated all the relevant example notebooks to reflect the new nomenclature.  We no longer represent each token with both a "char_span" and a "token_span" column. 
* `Analyze_Model_Outputs.ipynb` and `Analyze_Text.ipynb` still use `TokenSpanDtype`, but they only use it to store spans that both cover multiple tokens and are constrained to start and end on token boundaries.


text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -139,13 +139,13 @@ def char_span_to_arrow(char_span: CharSpanArray) -> pa.ExtensionArray:
     return pa.ExtensionArray.from_storage(typ, storage)
 
 
-def arrow_to_char_span(extension_array: pa.ExtensionArray) -> CharSpanArray:
+def arrow_to_char_span(extension_array: pa.ExtensionArray) -> SpanArray:
```
BryanCutler, line 142: change function name to `arrow_to_span` ?

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -24,12 +24,12 @@
 import numpy as np
 import pyarrow as pa
 
-from text_extensions_for_pandas.array import CharSpanArray, TensorArray, TokenSpanArray
+from text_extensions_for_pandas.array import SpanArray, TensorArray, TokenSpanArray
 
 
 class ArrowCharSpanType(pa.PyExtensionType):
```
BryanCutler, line 30: Might be a good idea to rename these as well to keep consistent. I could do that as a followup though.

text_extensions_for_pandas/array/arrow_conversion.py
BryanCutler: I can do this in followup as well

notebooks/Analyze_Model_Outputs.ipynb
```diff
@@ -95,8 +95,7 @@
        "  <thead>\n",
```
BryanCutler, line 95: <p><code>sentence</code>  references <code>TokenSpan</code>,  maybe remove the reference or briefly mention what it means?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/103/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

text_extensions_for_pandas/array/arrow_conversion.py
frreiss: Good catch! I missed that while doing search-and-replace. Fixed now.

text_extensions_for_pandas/array/arrow_conversion.py
frreiss: I've made this change too.

notebooks/Analyze_Model_Outputs.ipynb
frreiss: Yeah, that reference comes out of the blue in this version. I've removed the reference to `TokenSpan` here, since there is a detailed explanation of token-based spans a few cells later in the same notebook.

frreiss:
CI found a regression. Pushed a fix.

PR #106, opened 2020-09-04
Use TensorElement for single row of TensorArray

BryanCutler:
This wraps a single row np.ndarray from a `TensorArray` in a `TensorElement` class. Pandas does internal checks after getting single rows and fails if they are np.ndarray because it expects a length of 1, but gets something different. TensorArray and TensorElement also now implement the `__array__` interface which allows creating a numpy array automatically through numpy apis. For example, below will produce a np.ndarray of shape (3,) with dtype np.int64

```
arr = np.array(TensorArray([1, 2, 3]))
```

### DIFF
--- a/text_extensions_for_pandas/array/tensor.py
+++ b/text_extensions_for_pandas/array/tensor.py
@@ -138,13 +138,15 @@ class TensorArray(pd.api.extensions.ExtensionArray, TensorOpsMixin):
             if len(values) == 0:
                 self._tensor = np.array([])
             else:
-                values = [v.to_numpy() if isinstance(v, TensorElement) else v for v in values]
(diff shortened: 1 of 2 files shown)


BryanCutler:
@frreiss this fixed a lot of Pandas tests and most of setitem is enabled now. Still other tests need to be enabled, but I can circle back to those later. I think this captures basic functionality pretty well now.

frreiss (approved):
LGTM. This design change makes a lot of sense to me. The ExtensionArray mechanism bakes in an assumption that each element of the array can be represented by a Python class, and that the Python class is itself part of the extension library.

frreiss:
Looks like that stalled build is just Travis being Travis. Merging this PR.

BryanCutler:
Thanks @frreiss !

PR #117, opened 2020-09-10
Refactor TokenSpanArray base class

### LINKED ISSUE #91: Create BaseSpanArray class as common subclass to CharSpanArray and TokenSpanArray
Currently, TokenSpanArray is derived from CharSpanArray, but is not a proper instance of a CharSpanArray and is missing certain attributes. For example the test for TokenSpanArray in TestPandasMethods.test_equals fails because it uses the CharSpanArray.equals method and does not have attributes like `_equivalent_arrays`.

It would make the class structure clearer if common functionality was moved to an abstract BaseSpanArray class and specific functionality implemented in the concrete classes.

BryanCutler:
This change makes TokenSpanArray a true subclass of SpanArray to simplify code and allow TokenSpanArray to inherit all of SpanArray attributes. Also, replacing span_util.py with mixin classes for span additions.

fixes #91


BryanCutler:
@frreiss this is my shot at making TokenSpanArray a proper subclass of SpanArray - hopefully I'm not missing any major issues here. If you still prefer the current class hierarchy, then we don't have to go this route too.

frreiss (approved):
The general approach here looks good. Thanks for cleaning this up!

Issue #118, opened 2020-09-10
Symbol locations in generated docs don't match intended locations


frreiss:
The generated API docs show functions and classes as being located at the regions of the namespace where their source files reside. For example, the docs for the `Span` class reference that class as `text_extensions_for_pandas.array.char_span.Span` even though we export this class at the top level of our namespace as `text_extensions_for_pandas.Span`.

The locations of symbols in the API docs should match the locations where we intend users to access them from.

frreiss:
I am looking into this.

Issue #97, opened 2020-08-27
Expose the contents of `io` as multiple sub-packages


frreiss:
Currently, all the public functions under `text_extensions_for_pandas.io` are exposed in the top-level package of the project. This portion of the namespace is getting crowded, and some of the names are becoming confusing. For example, it's not clear from the name of the `make_tokens_and_features()` function that the function is part of the SpaCy integration.

We should expose the functions under `io` as a hierarchy of functions, not a flat collection. I recommend using the current location within the source tree to define the hierarchy. For example, `tp.make_tokens_and_features()` would become `tp.io.spacy.make_tokens_and_features()`.

FYI @BryanCutler

BryanCutler:
+1, having a hierarchy will make it clear what it is being used for.

frreiss:
Closed by #119

PR #119, opened 2020-09-11
Adjust package structure to get Sphinx working properly

frreiss:
This PR modifies the `__init__.py` files throughout the project so that Sphinx autodoc will correctly pick up the names of API functions. These changes address the problems I noted in #118, as well as getting the repository closer to a point where we can generate API docs from the https://readthedocs.org/ Github hooks.

Since I was mucking with package headers anyhow, I also made the naming changes to cover issue #97. The contents of the `io` directory are now exposed as a collection of sub-packages, one package for each library we integrate with. I've also given the `spanner` and `jupyter` sub-packages their own portions of the namespace, instead of having everything there imported to the top-level namespace. Now the only thing directly under `text_extensions_for_pandas` is the classes that comprise the extension types.

I've modified the notebooks to account for new subpackage names where needed.

I replaced our previous usage of `sphinx-apidoc` with a hard-coded `index.rst` file that pulls in all the public packages. That approach seems to be simpler and easier than the additional code changes we would need for `apidoc` to output a usable set of files. We may want to go in at a later date and re-enable `apidoc`. If we do re-enable it, we should invoke it from the `conf.py` file via the `sphinxcontrib.apidoc` extension (https://github.com/sphinx-contrib/apidoc). I've left a dependency on `sphinxcontrib-apidoc` in `config/dev_reqs.txt` to support such an eventual change.

For now, all the API documentation is ends up in single HTML file (plus some auxiliary files for indexes and searching). We can change that layout in the future by splitting `index.rst` into multiple files if needed.


text_extensions_for_pandas/array/__init__.py
```diff
@@ -20,11 +20,13 @@
...
 
 # For now we expose every symbol of the subpackages
-from text_extensions_for_pandas.array.span import *
-from text_extensions_for_pandas.array.token_span import *
-from text_extensions_for_pandas.array.tensor import *
-from text_extensions_for_pandas.array.accessor import *
+# from text_extensions_for_pandas.array.span import *
+# from text_extensions_for_pandas.array.token_span import *
+# from text_extensions_for_pandas.array.tensor import *
+# from text_extensions_for_pandas.array.accessor import *
```
BryanCutler, line 30: did you want to leave these commented out?

BryanCutler (approved):
LGTM, the module restructuring looks better now.

BryanCutler:
I'll merge now so we can make progress on the docs, the minior nit about the comments can be looked at later.

text_extensions_for_pandas/array/__init__.py
frreiss: Oops! I'll fix that in my next PR.

PR #128, opened 2020-09-25
Fixes for Python 3.6

BryanCutler:
Fixed imports causing problems with Python 3.6. Import in the form of `import text_extensions_for_pandas.io.bert as bert` end up causing a circular reference due to later imports also referencing `io` which has not been initialized yet. Using the form `from text_extensions_for_pandas.io import bert` avoid this.

### DIFF
--- a/text_extensions_for_pandas/io/test_bert.py
+++ b/text_extensions_for_pandas/io/test_bert.py
@@ -13,12 +13,14 @@
 #  limitations under the License.
 #
 
+import sys
 import unittest
 
 import pandas as pd
 import numpy as np
 import textwrap
 from typing import *
+import pytest
 
 # noinspection PyPackageRequirements
 from transformers import BertTokenizerFast, BertModel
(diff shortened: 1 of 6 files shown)


text_extensions_for_pandas/io/test_bert.py
```diff
@@ -32,6 +34,7 @@
 )
 
 
+#@pytest.mark.skipif(sys.version_info[:2] <= (3, 6), reason="BertModel segfault")
```
BryanCutler, line 37: I'm getting a segfault locally when constructing the `BertModel`, maybe an env thing so see what happens in travis

BryanCutler:
cc @kmh4321 @frreiss

text_extensions_for_pandas/io/test_bert.py
BryanCutler: yup, looks like it's just my env :confused:

kmh4321 (approved):
Looks great! I also encountered the circular imports causing this issue but your solution is better than my proposed solution of relative imports. thanks!

BryanCutler:
Great, thanks for checking @kmh4321 . This also adds Python 3.6 to travis CI to make sure we don't break the imports again. I'll go ahead and merge.

frreiss:
LGTM. I notice that this change set leaves `env.sh` unchanged, which means we will be developing with Python 3.7. I think that's ok for now, since Travis will be testing with 3.6.

PR #129, opened 2020-09-29
Update binder config

frreiss:
This PR improves the binder configuration. Some initialization steps would crash the container due to running out of disk space. I've moved those steps to image creation, where there is more disk available. I've also split out the larger, slower-to-install optional packages like `torch` into a separate requirements file, in case we want to have an environment build that is light on disk space. I've removed some extra dependencies from `dev_env.yml` that were causing `env.sh` to install TensorFlow 2.0, which messed up the `transformers` package. I've also adjusted some of the instructions in the notebooks.

I've left the binder link pointing at my branch for now, since that branch will change less often than master and Binder rebuilds its Docker images every time your target branch gets a new commit.


notebooks/Model_Training_with_BERT.ipynb
```diff
@@ -1420,7 +1420,7 @@
        "      <td>&lt;NA&gt;</td>\n",
```
BryanCutler, line 1420: <p>Whoa, I thought my 18min training was slow :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

BryanCutler, line 1420: <p>hmm, any idea why you got more true positives than me, just random?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

BryanCutler (approved):
LGTM, the binder nb run for me now.

notebooks/Model_Training_with_BERT.ipynb
frreiss: <p>Yes, the algorithm used is literally random. Well, randomized. I haven't found a way to fix the random seeds used.</p>

frreiss: <p>Usually these models are closer to 20 minutes. sklearn must have picked a bad initial solution.</p>

PR #130, opened 2020-09-29
Additional Binder-related updates.

frreiss:
This PR includes additional changes related to running our notebooks in a Binder environment.

I've reduced the memory footprint and running time of the model training notebook by adding in some Gaussian random projections on top of the embeddings. Currently, the notebook still runs out of memory on the Binder environment but it comes closer to running.

I've also fixed a typo in a file path in the tables notebook that was preventing it from running on Binder.


text_extensions_for_pandas/io/bert.py
```diff
@@ -118,14 +118,18 @@ def make_bert_tokens(target_text: str, tokenizer) -> pd.DataFrame:
...
     """
     Add BERT embeddings to a DataFrame of BERT tokens.
 
     :param df: DataFrame containing BERT tokens, as returned by
       :func:`make_bert_tokens` Must contain a column
       `input_id` containing token IDs.
     :param bert: PyTorch-based BERT model from the `transformers` library
+    :param overlap: (optional) how much overlap there should be between adjacent windows
+    :param non_overlap: (optional) how much non-overlapping content between the
+     overlapping regions there should be at the middle of each window?
```
BryanCutler, line 132: is the `non_overlap` param a rhetorical question? haha j/k
also the one below too

BryanCutler (approved):
LGTM. I was wondering if there was a way we could setup binder to have separate configs for the different notebooks? The overview notebook has minimal dependencies and would load quite a bit quicker. Not sure if that is at all possible though..

text_extensions_for_pandas/io/bert.py
frreiss: Yeah, window parameters are confusing. We should probably put an illustration into the docs at some point.

frreiss:
@BryanCutler Binder images are tied to the repository as far as I can see. I suppose we could create a separate branch for each notebook...

text_extensions_for_pandas/io/bert.py
BryanCutler: Well, I was just referring to the question mark at the end of the description..

PR #134, opened 2020-10-01
Remove TensorArray contiguous option

BryanCutler:
Removing the contiguous option in the TensorArray constructor. It's not needed, except for the Arrow conversion and is already forced there, and to be used properly would need to propagate when making a new TensorArray from an existing one.


BryanCutler:
I don't think this option has been used anywhere, so merging.

PR #138, opened 2020-10-09
Fix TensorArray aggregations that produce ndarray of objects

### LINKED ISSUE #124: Groupby+sum aggregate turns TensorArray into array of arrays
`sum` aggregates applied inside a `groupby` operation produce a TensorArray where the backing array is an array of arrays instead of a single n-dimensional array.

Code to reproduce:
```python
>>> import text_extensions_for_pandas as tp
    import numpy as np
    import pandas as pd

    df = pd.DataFrame({
        "a": ["foo", "bar"],
        "b": tp.TensorArray(np.array([[1, 2], [3, 4]]))
    })
    result = df.groupby("a").aggregate({"b": "sum"})
    print(result["b"].array)
```
Output:
```
array([array([3, 4]), array([1, 2])], dtype=object)
```

Expected output:
```
array([[3, 4], [1, 2]], dtype=int)
```

BryanCutler:
`TensorArray` aggregations were producing a ndarray of `TensorArray`s for each group. This changes aggs to produce a `TensorElement` as a scalar, then allow construction of a new `TensorArray` using an ndarray of the resulting `TensorElement` objects.

Fixes #124

### DIFF
--- a/text_extensions_for_pandas/array/tensor.py
+++ b/text_extensions_for_pandas/array/tensor.py
@@ -140,7 +140,11 @@ class TensorArray(pd.api.extensions.ExtensionArray, TensorOpsMixin):
         :param values: A `numpy.ndarray` or sequence of `numpy.ndarray`s of equal shape.
         """
         if isinstance(values, np.ndarray):
-            self._tensor = values
+            if values.dtype.type is np.object_ and len(values) > 0 and \
+                    isinstance(values[0], TensorElement):
+                self._tensor = np.array([np.asarray(v) for v in values])
+            else:
+                self._tensor = values
         elif isinstance(values, Sequence):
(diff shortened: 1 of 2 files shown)


BryanCutler:
also checked the print out is correct:
```
In [3]:     df = pd.DataFrame({ 
   ...:         "a": ["foo", "bar"], 
   ...:         "b": tp.TensorArray(np.array([[1, 2], [3, 4]])) 
   ...:     }) 
   ...:     result = df.groupby("a").aggregate({"b": "sum"}) 
   ...:     print(repr(result["b"].array)) 
   ...:                                                                                                                      
array([[3, 4],
       [1, 2]])
```
will go ahead and merge since tests are green, cc @frreiss

PR #139, opened 2020-10-14
Minor change: Add jupyterlab-git to default environment

frreiss:
This one-line config file change adds the git plugin to the JupyterLab setup that `env.sh` creates.


frreiss:
Will merge this when CI passes.

PR #140, opened 2020-10-14
Fix CI issues

frreiss:
This PR should hopefully fix the issues we've been having with CI builds hanging.

I've added proper argument-handling code to `env.sh`, and I've fixed a bug that caused that script to hang if the `PANDAS_VERSION` environment variable is set.

`env.sh` now looks for an environment variable `PYTHON_VERSION` (there's also a command-line argument `--python_version` for non-Travis use) and installs the specified version of Python.

I've updated all the Travis builds to use the `PYTHON_VERSION` environment variable, and I've added an additional build for Python 3.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,19 @@ language: python
 
 jobs:
   include:
-    - python: "3.6"
+    - name: "Python 3.6"
+      python: "3.6"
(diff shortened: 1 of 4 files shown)


BryanCutler (approved):
LGTM

PR #141, opened 2020-10-15
Prep for second alpha release

frreiss:
This PR increments the version number on master to `0.1a2` and adds instructions for creating releases. Once this PR is merged, I'll test out those instructions with a second alpha release.

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -1,0 +1,50 @@
+
+# Release process
+
+Steps to release a new version:
+
+1. Ensure all regression tests are passing on all supported versions of Python
+   and Pandas. The current versions are:
+   * Python: 3.6, 3.7, 3.8
+   * Pandas: 1.0.x, 1.1.x
+
+1. Activate your Text Extensions for Pandas build environment (usually called 
+   `pd`)
+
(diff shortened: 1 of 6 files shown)


docs/release.md
```diff
@@ -0,0 +1,50 @@
...
+   * Python: 3.6, 3.7, 3.8
+   * Pandas: 1.0.x, 1.1.x
+
+1. Activate your Text Extensions for Pandas build environment (usually called 
+   `pd`)
+
+1. Ensure that all the notebooks under the `notebooks` directory run and
+   produce substantially the same output as before. This step only needs to be
+   done with our primary versions of Python and Pandas at the time of the
+   release.
```
BryanCutler, line 17: Is it possible to test this in CI?

BryanCutler (approved):
LGTM

docs/release.md
BryanCutler: Looks like the best option is to convert notebook to a script with `nbconvert` package, then run the script. Maybe too much for CI, but we could make a bash script to run them all and launch it manually, wdyt? We should still run each notebook to check the output before release, but this might help to flush out any api changes..

PR #142, opened 2020-10-17
Add TensorArray numpy properties for dtype, ndims and shape

BryanCutler:
Adding convenience methods to get numpy properties, so it's not necessary to convert using `to_numpy()`.


BryanCutler:
I'll merge this, we can make changes in the api later if needed.

PR #144, opened 2020-10-24
Enable parquet saving since supported in Arrow 2.0.0

BryanCutler:


BryanCutler:
confirmed it runs in the Python 3.8 CI job

Issue #54, opened 2020-07-14
Extend Watson NLU entities translation to include individual entity spans


frreiss:
By default, the `entities` tool in Watson Natural Language Understanding only outputs aggregate information about entity mentions in the document. Our Pandas translation currently surfaces this information as a dataframe with one row per aggregate entity:
![image](https://user-images.githubusercontent.com/12436991/87449912-ab7a7000-c5b2-11ea-8e2f-0c0d1d40ea26.png)

The `entities` tool has an option called `mentions` that is off by default, but if you turn this option on...

```python
entities=EntitiesOptions(mentions=True)
```

...the returned JSON value will have detailed information about specific entity mentions, including span information:
```json
{
      "type": "Person",
      "text": "Lancelot",
      "sentiment": {
        "score": 0.835873,
        "label": "positive"
      },
      "relevance": 0.678523,
      "mentions": [
        {
          "text": "Lancelot",
          "location": [
            1393,
            1401
          ],
          "confidence": 0.614782
        },
        {
          "text": "Lancelot",
          "location": [
            2884,
            2892
          ],
          "confidence": 0.99065
        },
        {
          "text": "Lancelot",
          "location": [
            2894,
            2902
          ],
          "confidence": 0.984138
        },
        {
          "text": "Lancelot",
          "location": [
            4253,
            4261
          ],
          "confidence": 0.989195
        },
        {
          "text": "Lancelot",
          "location": [
            4759,
            4767
          ],
          "confidence": 0.977934
        }
      ],
      "count": 5,
      "confidence": 1
    },
```

We should extend our `watson_nlu_parse_response()` function so that it detects the presence of this detailed information and produces a dataframe with one row per mention (as opposed to one row per aggregate) when the detailed information is available.

frreiss:
Closed by #146

PR #146, opened 2020-11-06
Add code to convert NLU entity mentions information to a DataFrame

frreiss:
This PR implements the feature described in #54.

I've added code to `nlu.py` to unroll the nested information that NLU returns about entity mentions into a DataFrame with one row per entity mention. After my modifications, the `parse_response()` method returns a dictionary with an additional key "entity_mentions" that points to the new DataFrame. If the user didn't tell NLU to extract entity mentions, this DataFrame will be empty but will still be present.

I added some new tests to cover the new functionality. I also updated the main NLU-related notebook with some Markdown to explain what's stored under the "entity_mentions" key in the return value of `parse_response()`.

### DIFF
--- a/text_extensions_for_pandas/io/watson/nlu.py
+++ b/text_extensions_for_pandas/io/watson/nlu.py
@@ -34,7 +34,7 @@ import numpy as np
 import pandas as pd
 import pyarrow as pa
 
-from text_extensions_for_pandas.array.span import SpanArray
+from text_extensions_for_pandas.array.span import SpanArray, Span
 from text_extensions_for_pandas.array.token_span import TokenSpanArray
 from text_extensions_for_pandas.io.watson import util
 from text_extensions_for_pandas.spanner import contain_join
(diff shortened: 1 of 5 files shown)


text_extensions_for_pandas/io/watson/nlu.py
```diff
@@ -337,14 +365,65 @@ def _make_relations_dataframe_zero_copy(relations):
...
+        # No mentions to unroll. Return an empty DataFrame.
+        return util.apply_schema(
+            pd.DataFrame(columns=[e[0] for e in _entity_mentions_schema]),
+            _entity_mentions_schema, apply_standard_schema)
+    if original_text is None:
+        raise ValueError(
+            "Unable to construct target text for converting entity mentions to spans")
+    # Explode out the nested relations containing entity location information.
+    # If there was a version of DataFrame.explode() that could handle structs,
+    # we would be able to vectorize this operation.
```
BryanCutler, line 390: Could you use `util.make_table` which flattens the struct columns? That is pretty much the reason I made that function so that you could work with non-nested columns in a vectorized way after flattening.

BryanCutler (commented):
Looks good, thanks for doing this Fred. I must have missed that option and the issue being created or I would have taken this on earlier. I just had one suggestion about flattening the struct columns in Arrow with `util.make_table` instead of unrolling 1 record at a time.

text_extensions_for_pandas/io/watson/nlu.py
frreiss: I tried to, but it wasn't working for me. How about if we merge this current change set, and then you can take a look at your leisure and see if you can remove that loop from the Python code?

BryanCutler: sure, will do

PR #152, opened 2020-11-24
Vectorize Watson Entity Mentions Processing

BryanCutler:


text_extensions_for_pandas/io/watson/test_nlu.py
```diff
@@ -247,10 +247,10 @@ def test_entity_mentions(self):
         self.assertEqual(
             repr(result["entity_mentions"]),
             textwrap.dedent("""\
-                       type           text                         span  confidence
-                0    Person  Steven Wright    [64, 77): 'Steven Wright'    0.998251
-                1    Person  Steven Wright  [177, 190): 'Steven Wright'    0.998251
-                2  Location         Alaska         [138, 144): 'Alaska'    0.999498""")
+                       type                         span  confidence
```
BryanCutler, line 250: I removed the text column because it's redundant with the span, similar to other DataFrames with spans

text_extensions_for_pandas/io/watson/nlu.py
```diff
@@ -365,54 +346,67 @@ def _make_relations_dataframe_zero_copy(relations):
...
+        pdf_mentions = table_mentions.to_pandas()
+        pdf_mentions["span"] = char_span
+
+        # Align index of parent entities DataFrame with flattened DataFrame and ffill values
+        mention_offsets = mention_arrays[0].offsets.to_numpy()
+        pdf_parent = pdf.set_index(mention_offsets[:-1])
+        pdf_parent = pdf_parent.reindex(pdf_mentions.index, method="ffill")
+
+        # Add columns from entities parent DataFrame
+        pdf_mentions["type"] = pdf_parent["type"]
```
BryanCutler, line 400: I seems like we only want "type" column from the entities df, unless "relevance" is useful too?

BryanCutler:
@frreiss please take a look when you can, thanks!

text_extensions_for_pandas/io/watson/test_nlu.py
frreiss: In this case, the `text` column actually serves a purpose. That column is part of the join key for joining entity mentions in the entity mentions DataFrame with the document-level information (sentiment, emotion, and document-level confidence) in the entities DataFrame. The whitespace and capitalization of the value in the `text` column could potentially be different from the covered text of the span.

The current implementation in NLU tends to treat multiple mentions of the same entity with slightly different surface forms as completely different entities, but I don't think we can assume that will always be the case. NLU ought to treat `John Smith`, `John  Smith` and `john Smith` as the same entity for the purposes of computing things like sentiment and emotion.

How about if we keep this column in the entity mentions DataFrame, but rename the column from "text" to something like "name" across both entity-related DataFrames?

text_extensions_for_pandas/io/watson/nlu.py
frreiss: The set of columns from the "entities" result that we include in the "entity_mentions" DataFrame is something of a judgement call. On one hand, including additional fields would save the user the trouble of calling `pd.merge()` on the two DataFrames. On the other hand, the properties in the parent entity record are associated with document-level entities, and we don't want to create the impression of a direct connection to individual mentions. There's also the issue of cluttering the entity mentions DataFrame with extra columns.

Here's an example. If you run this code:
```python
natural_language_understanding.analyze(
    text="I hate John Smith. You love John Smith.",,
    return_analyzed_text=True,
    features=nlu.Features(
        entities=nlu.EntitiesOptions(sentiment=True, mentions=True, emotion=True)
    )).get_result()
```
you get back this result:
```
{'usage': {'text_units': 1, 'text_characters': 39, 'features': 1},
 'language': 'en',
 'entities': [{'type': 'Person',
   'text': 'John Smith',
   'sentiment': {'score': -0.296829, 'mixed': '1', 'label': 'negative'},
   'relevance': 0.9999,
   'mentions': [{'text': 'John Smith',
     'location': [7, 17],
     'confidence': 0.995878},
    {'text': 'John Smith', 'location': [28, 38], 'confidence': 0.992923}],
   'emotion': {'sadness': 0.112204,
    'joy': 0.289177,
    'fear': 0.061407,
    'disgust': 0.134739,
    'anger': 0.409954},
   'count': 2,
   'confidence': 0.999971}],
 'analyzed_text': 'I hate John Smith. You love John Smith.'}
```

Note how there are two mentions of "John Smith" with very different local sentiment, but NLU associates both of these with the same document-level sentiment information (`sentiment': {'score': -0.296829, 'mixed': '1', 'label': 'negative'}`) and document-level emotion information (`'emotion': {'sadness': 0.112204,...`). Also, the response includes **three** confidence values: One for each of the entity mentions and one for the document-level entity "John Smith".

frreiss (approved):
This looks great! I will use the lessons learned from reviewing this code in my future dealings with pyarrow :-)

I've made some minor clarifications about NLU semantics in my inline comments.

One additional request: While reviewing this PR I noticed that there is an additional field, "emotion", that can appear in a sentiment record. Your new code should pick that field up as is, but it would be good to test that behavior. Would you mind adding an "emotion" field to the input for the regression test? You can find an example value in my first inline comment here.

text_extensions_for_pandas/io/watson/test_nlu.py
BryanCutler: Ah ok, that makes sense to keep the `text` column then. I think we should keep it as `text` though and not rename to be consistent with the Watson response, that ok with you?

text_extensions_for_pandas/io/watson/nlu.py
BryanCutler: Yeah, sounds good. Thanks for the great explanation! I'll include this in the tests.

BryanCutler:
Thanks @frreiss , I'll work on an update.

text_extensions_for_pandas/io/watson/test_nlu.py
frreiss: Works for me. Thanks!

BryanCutler:
Waiting on #155 to get CI back up

BryanCutler:
I think I addressed feedback and tests now pass, going to merge.

PR #154, opened 2020-12-03
Add github workflow file for CI

BryanCutler:


BryanCutler:
I think this needs to be committed to run the action, going to try it out.

Issue #151, opened 2020-11-23
2+ dimensional tensors of timestamps crash pd.Series.repr()


frreiss:
**Code to reproduce:**
```python
import text_extensions_for_pandas as tp
import pandas as pd
import numpy as np

times = pd.date_range('2018-01-01', periods=5, freq='H').to_numpy()
times_repeated = np.tile(times, (3, 1))
times_array = tp.TensorArray(times_repeated)
times_series = pd.Series(times_array)
print(repr(times_series))
```

**Expected result:** Display a 3x5 matrix of timestamps

**Actual result:** Crash from inside a Pandas routine that should only be called for 1-D arrays of timestamps. Stack trace follows.
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-b6b4c1b73ef2> in <module>
      7 times_array = tp.TensorArray(times_repeated)
      8 times_series = pd.Series(times_array)
----> 9 repr(times_series)

~/pd/cn-update/env/lib/python3.7/site-packages/pandas/core/series.py in __repr__(self)
   1331             min_rows=min_rows,
   1332             max_rows=max_rows,
-> 1333             length=show_dimensions,
   1334         )
   1335         result = buf.getvalue()

~/pd/cn-update/env/lib/python3.7/site-packages/pandas/core/series.py in to_string(self, buf, na_rep, float_format, header, index, length, dtype, name, max_rows, min_rows)
   1396             max_rows=max_rows,
   1397         )
-> 1398         result = formatter.to_string()
   1399 
   1400         # catch contract violations

~/pd/cn-update/env/lib/python3.7/site-packages/pandas/io/formats/format.py in to_string(self)
    356 
    357         fmt_index, have_header = self._get_formatted_index()
--> 358         fmt_values = self._get_formatted_values()
    359 
    360         if self.truncate_v:

~/pd/cn-update/env/lib/python3.7/site-packages/pandas/io/formats/format.py in _get_formatted_values(self)
    345             None,
    346             float_format=self.float_format,
--> 347             na_rep=self.na_rep,
    348         )
    349 

~/pd/cn-update/env/lib/python3.7/site-packages/pandas/io/formats/format.py in format_array(values, formatter, float_format, na_rep, digits, space, justify, decimal, leading_space, quoting)
   1177     )
   1178 
-> 1179     return fmt_obj.get_result()
   1180 
   1181 

[...]

PR #157, opened 2020-12-29
Fix test failures with Pandas 1.2.0

### LINKED ISSUE #158: Support Pandas version 1.2.0
Currently there are test failures when using Pandas 1.2.0. Need to resolve these to fully support Pandas 1.2.0 and above.

BryanCutler:
There are a number of test failures when using Pandas 1.2.0.

* The method `pandas.core.ops._get_op_name()` has been removed in Pandas >=1.2.0. This was being used when building `TensorArray` ops, and is replaced with similar usage in `ExtensionScalarOpsMixin`.
* `TensorArray` should return it's instance when `astype()` called with same type
* `TensorArray._contains_` needs to have proper implementation of `any()`
* `repr()` is broken when using float values of ndims > 2

Fixes #158

### DIFF
--- a/text_extensions_for_pandas/io/watson/test_tables.py
+++ b/text_extensions_for_pandas/io/watson/test_tables.py
@@ -13,10 +13,11 @@
 #  limitations under the License.
 #
 
+from distutils.version import LooseVersion
 import json
-import os
-import textwrap
 import unittest
+import pandas as pd
+import pytest
 
 from text_extensions_for_pandas.io.watson.tables import *
 
@@ -445,6 +446,8 @@ class TestTables(unittest.TestCase):
 15        Total tax rate  \
 """)
 
+    @pytest.mark.skipif(LooseVersion(pd.__version__) >= LooseVersion("1.2.0"),
+                        reason="TODO: Rank col gets converted to float")
     def test_make_table(self):
         double_header_table = make_table(parse_response(self.responses_dict["double_header_table"]))
         self.assertEqual(repr(double_header_table), """\
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -19,14 +19,18 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
-        # Test against Pandas 1.0 and latest version
-        pandas-version: ["1.0.*", ""]
+        # Test against Pandas 1.0, 1.1, and latest version
+        pandas-version: ["1.0.*", "1.1.*", ""]
         exclude:
-          # Only run one test with Pandas 1.0.x and Python 3.7, exclude others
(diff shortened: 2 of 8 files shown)


BryanCutler:
There seem to be a number of different issue with Pandas, I'll look into fixing them here too.

frreiss:
@BryanCutler any idea what's causing that test failure on the Python 3.7 build (https://github.com/CODAIT/text-extensions-for-pandas/pull/157/checks?check_run_id=1620791768)?

BryanCutler:
I fixed most of the issues with pandas 1.2.0 and TensorArray, except there is a regression with `repr()` with floats, similar to #151. There has been some back and forth with my fix upstream for that one, so maybe I can fix both. Unfortunately, I don't know of a workaround, so we either have to leave it as a known issue or cap the supported pandas version < 1.2.0.  I'm looking into the other issues now.

text_extensions_for_pandas/io/watson/test_tables.py
```diff
@@ -445,6 +446,8 @@ def test_make_exploded_df(self):
 15        Total tax rate  \
 """)
 
+    @pytest.mark.skipif(LooseVersion(pd.__version__) >= LooseVersion("1.2.0"),
+                        reason="TODO: Rank col gets converted to float")
```
BryanCutler, line 450: Not sure why this is different now, but I think it's safe to skip and I can make an issue to follow up with later

BryanCutler:
@frreiss this failure https://github.com/CODAIT/text-extensions-for-pandas/pull/157/checks?check_run_id=1646927381#step:5:1438 in `bert.align_bert_tokens_to_corpus_tokens()` seems a bit strange. Not sure why this would suddenly change, unless they added some optimization that wasn't complete. I can look into it further, but letting you know in case you have some idea what's going on.

BryanCutler:
Ok, I think I found why the above error is happening and looks to be a bug. Pandas attempts to do a cython aggregation and then fallback to a vanilla agg if there is an error, but checks the error message improperly for this case. I should be able to file a bug report and do a PR for the fix.

frreiss:
Thanks for tracking down the root cause of that problem, @BryanCutler !

BryanCutler:
PR for the agg fix https://github.com/pandas-dev/pandas/pull/38982

BryanCutler:
@frreiss I capped the upper version of Pandas to < 1.2.0 because not being able to display a Series with TensorArray of floats is pretty major and I don't see a workaround. I'll keep working on the upstream fix, but what are your thoughts on merging this?

text_extensions_for_pandas/io/watson/test_tables.py
frreiss: Works for me.

frreiss:
I think merging is the best way forward for now. We may want to post a new release just to get that pandas<1.2.0 constraint into our requirements.txt.

PR #165, opened 2021-01-20
Fixed issue in notebook with TenosrArray bool selection

BryanCutler:


text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -419,6 +419,25 @@ def test_numpy_properties(self):
...
+        npt.assert_array_equal(result, expected)
+
+        # Test Series of TensorDType selection with numpy array
+        s = pd.Series(data)
+        result = s[np.asarray(sel)]
+        npt.assert_array_equal(result, expected)
+
+        # Test Series  of TensorDType selection with TensorArray
+        # Currently fails due to Pandas not recognizing as bool index GH#162
+        with self.assertRaises(Exception):
```
BryanCutler, line 438: Looking into a fix upstream, I think pandas should be able to recognize as a bool index

notebooks/Text_Extensions_for_Pandas_Overview.ipynb
```diff
@@ -1811,7 +1817,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.7"
+   "version": "3.8.6"
```
frreiss, line 1820: I think your local copy of JupyterLab isn't the one that `env.sh` creates. That script installs Python 3.7.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -35,6 +35,7 @@ class TensorDtype(pd.api.extensions.ExtensionDtype):
     """
     Pandas data type for a column of tensors with the same shape.
     """
+    base = None
```
frreiss, line 38: What does this do? Should this change be repeated on the dtypes for span arrays?

text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -419,6 +419,57 @@ def test_numpy_properties(self):
         self.assertEqual(arr.numpy_shape, data.shape)
         self.assertEqual(arr.numpy_dtype, data.dtype)
 
+    def test_bool_tensor_selection(self):
+        data = TensorArray([[1, 2], [3, 4], [5, 6]])
+        sel = TensorArray([True, False, True])
+        expected = np.array([[1, 2], [5, 6]])
+
+        # Test TensorArray.__getitem__ with TensorArray
+        result = data[sel]
```
frreiss, line 428: You should probably test what happens when `sel` is a 2D tensor. Is that case supposed to return an error, or is it supposed to perform numpy indexing?

frreiss (approved):
Looks good to me. Some minor comments inline.

notebooks/Text_Extensions_for_Pandas_Overview.ipynb
BryanCutler: Oops, must have been a diff env. I'll fix it up and re-run

text_extensions_for_pandas/array/tensor.py
BryanCutler: I'm not sure, I can't find any documentation for it. It's in some extension types and only set to something for a couple. Without it, there is an error in a cython function when using integer as index. I'll followup with a question to the mailing list and try to add document or a fix upstream.

text_extensions_for_pandas/array/test_tensor.py
BryanCutler: Yeah, good idea

BryanCutler:
Thanks @frreiss, I'll work on an update then merge.

PR #167, opened 2021-01-22
Add "first" reduction type and NA value to spans

frreiss:
This PR adds a "first" reduction (aka aggregate) to `SpanArray` and `TokenSpanArray`. I've also implemented the `na_value()` hook on the associated dtype classes, so that empty groups in a `groupby` expression will get a reasonable value.

There is still some work to be done in handling the case where someone tries to roll up a list of `Span` objects, some of which are null values with null target text and some of which aren't, into a single `SpanArray`. I think that problem is best handled in a follow-on PR.

### DIFF
--- a/text_extensions_for_pandas/array/token_span.py
+++ b/text_extensions_for_pandas/array/token_span.py
@@ -217,6 +217,15 @@ class TokenSpanDtype(SpanDtype):
         """:return: A string representation of the dtype."""
         return "TokenSpanDtype"
 
+    @property
+    def na_value(self) -> object:
+        """
+        See docstring in `ExtensionDType` class in `pandas/core/dtypes/base.py`
+        for information about this method.
+        """
+        empty_span_array = SpanArray(None, [], [])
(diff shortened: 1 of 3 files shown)


text_extensions_for_pandas/array/token_span.py
```diff
@@ -218,6 +218,15 @@ def name(self) -> str:
...
         return "TokenSpanDtype"
 
+    @property
+    def na_value(self) -> object:
+        """
+        See docstring in `ExtensionDType` class in `pandas/core/dtypes/base.py`
+        for information about this method.
+        """
+        empty_span_array = SpanArray(None, [], [])
+        return TokenSpan(empty_span_array, Span.NULL_OFFSET_VALUE, Span.NULL_OFFSET_VALUE)
```
BryanCutler, line 228: As a followup, it would be good to use these values in an array to check that they don't cause problems.

BryanCutler (approved):
LGTM

Issue #73, opened 2020-08-05
Extend CharSpanArray and TokenSpanArray to support multiple documents


frreiss:
The current implementation of `CharSpanArray` and `TokenSpanArray` only allows a single target text for all of the spans in a given array. This restriction is fine as long as all the spans in a given Dataframe come from a single document, but it complicates use cases involving combining information from multiple documents in a single Dataframe. Currently the only way to have spans from multiple documents in a series is to convert `CharSpanArray`/`TokenSpanArray` arrays into arrays of type `Object` containing individual `CharSpan` and `TokenSpan` objects.

We should extend our span array types to allow for multiple target texts per array. Key challenges to address:
* Memory-efficient representation for the span data
* Clean semantics for span comparison across documents. What happens if two documents in a corpus have the same text?
* Efficient implementations of the span operations under `text_extensions_for_pandas.spanner` with multiple target texts
* Serialization/deserialization to/from Arrow and Feather format

frreiss:
WIP PR at #170

frreiss:
Closed by #170

frreiss:
Correction: #170 implements everything but updated Arrow support. Opened #179 to cover that work.

PR #170, opened 2021-01-26
Add multi-document support for span arrays

frreiss:
***Work in progress, do not merge yet.***

This PR contains the first step towards implementing #73. There's quite a bit of additional work to be done, but I'm sharing this WIP PR to get some feedback on the current design.

Here's what has been implemented so far:
* Added a new class, `StringTable`, for efficiently managing the connection between document texts and spans.
* Modified the `SpanArray` class to support mixing spans from multiple documents.
* The existing unit tests in `test_span.py` pass.

Major things still remaining:
* Modify the `TokenSpanArray` class to also support multiple documents
* Update the Arrow integration to support multiple documents per array
* Update the algorithms under the `spanner` package to support multiple documents per span array
* Get all existing regression tests to pass
* Add new regression tests for multi-document support
* Update the example notebooks as needed


frreiss:
@BryanCutler can you have a look at this design for multi-doc support and see if you notice any gotchas on either Arrow integration or passing the full Pandas ExtensionArray suite?

BryanCutler (commented):
I did a quick look and seems like a good approach to me. It shouldn't be a problem to update the arrow conversion to support it. I'll do a more thorough review soon.

frreiss:
Merged with changes from master, found a bunch of regressions, and fixed them.

frreiss:
@BryanCutler how would you recommend we encode `SpanArray`s with multiple target texts in Arrow? Should we directly serialize the string table and an array of integer offsets into the table? Should we use an Arrow Dictionary array? Something else?

BryanCutler:
I was thinking the best thing to do would be a DictionaryArray, but I'm not clear on how that would work with Pandas or if it's even implemented, so I'll have to try it out. If not, we could do something different, but might not be a clean as a dictionary.

BryanCutler:
@frreiss I did a preliminary test of using a nested DictionaryArray through saving/loading a feather file and it works, so we should definitely go that route. It also might work that we could use the same DictionaryArray across different columns and only write to the file once, to avoid duplicating the document text. I'll have test more for that, but in the meantime we should proceed. Do you want to finish up with this PR and then I can follow with the Arrow changes?

frreiss:
Makes sense, @BryanCutler. If we'll be using Arrow's native dictionary coding for serialization, do you think it would make sense to replace the guts of the `StringTable` class with Arrow's dictionary data structure?

BryanCutler:
>Makes sense, @BryanCutler. If we'll be using Arrow's native dictionary coding for serialization, do you think it would make sense to replace the guts of the StringTable class with Arrow's dictionary data structure?

It might be a good idea, that would make it so there isn't a need to copy documents when doing IO. Basically you would just need to change `_id_to_str` to a `pa.Array` of type string. Also, I think a numpy array of strings would work fine too, if you don't want to add an arrow dependency there.

frreiss:
Updated token-based spans to support multiple documents. This change required refactoring of the previous changes to character-based spans.

In the process, I changed the API for creating span arrays from a constructor call to a factory method. That is, `SpanArray.create()` instead of `SpanArray()`.

Updated Feather support is still **not** implemented. I've disabled the tests of Feather for now.

All the other regression tests are now working.

Still need to update the notebooks to reflect the constructor changes.

BryanCutler:
Sounds good @frreiss , as soon as this is merged I'll work on the arrow changes and enable the tests again.

frreiss:
Updated the notebooks and fixed a few additional bugs.

frreiss:
@BryanCutler I think these changes are about ready to merge, with the exception of revamping the Arrow/Feather support. Can you review these changes, please?

text_extensions_for_pandas/array/token_span.py
```diff
@@ -267,49 +296,96 @@ class TokenSpanArray(SpanArray, TokenSpanOpMixin):
...
                             f"Supported types are Series, ndarray and List[int].")
         if not isinstance(end_tokens, (pd.Series, np.ndarray, list)):
             raise TypeError(f"end_tokens is of unsupported type {type(end_tokens)}. "
                             f"Supported types are Series, ndarray and List[int].")
 
-        super().__init__(tokens.target_text, tokens.begin, tokens.end)
+        if isinstance(tokens, SpanArray):
+            if not tokens.is_single_document:
+                raise ValueError(f"Token spans come from more than one document.")
+            token_table = TokenTable.create_single(tokens)  # type: ThingTable
```
BryanCutler, line 325: Why do we need to create a table for tokens? Since the token positions are indices in the SpanArray, aren't the multiple docs already handled?

text_extensions_for_pandas/array/token_table.py
```diff
@@ -0,0 +1,84 @@
...
+# by span array types to store target document texts and their tokenizations.
+#
+
+from typing import *
+
+from text_extensions_for_pandas import SpanArray
+from text_extensions_for_pandas.array.thing_table import ThingTable
+
+
+class _BoxedSpanArray:
```
BryanCutler, line 31: Is the reason we need this because we are putting tokens as a `SpanArray` in here?

BryanCutler (commented):
@frreiss I have a few questions on some of the high-level designs first before I do a more detailed review.

text_extensions_for_pandas/array/token_table.py
frreiss: Yup. The `__eq__` method on an ExtensionArray returns a mask. Python dictionaries don't like that.

text_extensions_for_pandas/array/token_span.py
frreiss: Each span in the `TokenSpanArray` could be from a different tokenization of a different document. For example, you might mix spans over the SpaCy tokenization of document 1, spans over the BERT tokenization of document 1, and spans over the SpaCy documentation of document 2 in the same array. 

It's not really practical in this situation to maintain a single `SpanArray` to cover the entire universe of tokens. You don't know what documents exist in the current Python process. The implementation of `TokenSpanArray` here uses a separate `SpanArray` to represent a single tokenization of a single document. Each `TokenSpan` is tagged with a specific underlying tokenization.

The main purpose of the table is to track the set of distinct tokenizations. Keeping track of that set lets some operations run in constant time:
* detecting that every span in an array (or a slice of an array) is over the same tokens
* identifying all the distinct tokenizations present in an array of spans (for serialization)
* detecting that two copies of the same tokenization (i.e. different arrays, same spans) are the same tokenization

An alternative approach would be to store an array of pointers to `SpanArrays` (i.e. what is currently computed on-demand under the property `TokenSpan.tokens`) instead of using a table of distinct tokenizations. That approach would simplify the code at the expense of making some operations asymptotically slower. I might make that change; haven't decided yet.

text_extensions_for_pandas/array/span.py
```diff
@@ -290,20 +340,28 @@ class SpanArray(pd.api.extensions.ExtensionArray, SpanOpMixin):
     are character offsets into the target text.
     """
 
-    def __init__(self, text: str,
-                 begins: Union[pd.Series, np.ndarray, Sequence[int]],
-                 ends: Union[pd.Series, np.ndarray, Sequence[int]]):
```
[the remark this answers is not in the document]
frreiss: Let me see if I can work out a way to get back to a single constructor.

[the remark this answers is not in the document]
frreiss: Latest change set switches back to a conventional constructor.

text_extensions_for_pandas/array/span.py
```diff
@@ -314,27 +372,84 @@ def __init__(self, text: str,
...
+        self._string_table = string_table
+        self._text_ids = text_ids
+        self._begins = begins
+        self._ends = ends
+
+        # Reset all cached data.
+        self.increment_version()
+
+    @classmethod
+    def _construct(cls,
```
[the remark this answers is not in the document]
frreiss: Latest change set gets rid of this method.

text_extensions_for_pandas/array/token_span.py
frreiss: Went ahead and got rid of the token table. Now `TokenSpanArray` uses an array of pointers to `SpanArray` objects representing the tokenizations.

A few wrinkles with this new implementation: Some methods generate a `TokenTable` on demand, and I had to jump through hoops in a few places to keep NumPy from converting an array of `SpanArray`s into a 2D `ndarray`.

frreiss:
Pushed changes that simplify the way `TokenSpanArray` manages multiple tokenizations and restore the conventional constructor APIs. @BryanCutler can you please have a look when you get a chance?

frreiss:
Pushed some additional changes to HTML rendering so that spans will render correctly in JupyterLab dark mode and in VSCode notebooks.

text_extensions_for_pandas/array/span.py
```diff
@@ -382,43 +457,60 @@ def __getitem__(self, item) -> Union[Span, "SpanArray"]:
...
+        def _is_sequence_of_spans(seq: Any):
+            if isinstance(seq, SpanArray):
+                return True
+            if not isinstance(seq, (collections.abc.Sequence, np.ndarray)):
+                return False
+            else:
+                # For other sequences, check for everything being Span or None
+                for elem in seq:
+                    if elem is not None and not isinstance(elem, Span):
+                        return False
```
BryanCutler, line 484: Just a suggestion, what is here is fine too

```suggestion
                if any(elem is not None and not isinstance(elem, Span) for elem in seq):
                    return False
```

text_extensions_for_pandas/array/span.py
```diff
@@ -729,16 +836,64 @@ def make_array(cls, o) -> "SpanArray":
...
+         the regions of these documents that the spans cover.
         """
-        return self._text
+        return self._string_table.ids_to_things(self._text_ids)
+
+    @memoized_property
+    def document_text(self) -> Union[str, None]:
+        """
+        :return: if all spans in this array cover the same document, text of that
+         document. Returns None if the array is emptoy or if teh Spans in this
```
BryanCutler, line 856: sp: emptoy, teh

text_extensions_for_pandas/array/span.py
```diff
@@ -729,16 +836,64 @@ def make_array(cls, o) -> "SpanArray":
...
+          array cover more than one document.
+        """
+        if len(self._text_ids) == 0:
+            return None
+        if not self.is_single_document:
+            return None
```
BryanCutler, line 862: could you return a list of documents for this case? or always return a list, even for one document?

BryanCutler: I guess that would be the same as `target_text`, but just seems a little strange to return `None` for multiple documents. would it make sense to have
```
@memoized_property
def num_documents()

def document_text(i)
```
to return the `i`th document?

text_extensions_for_pandas/array/span.py
```diff
@@ -878,6 +1078,33 @@ def repr_html_show_offsets(self):
...
+
+    def _clear_cached_properties(self) -> None:
+        """
+        Remove cached values of memoized properties to reflect changes to the
+        data on which they are based.
+        """
+        for n in self._cached_property_names():
+            attr_name = "_{0}".format(n)
+            if hasattr(self, attr_name):
+                delattr(self, attr_name)
```
BryanCutler, line 1106: Hmm, this kind of sucks you have to do this to clear it.  So you remove the attr, does it need to be added back?

BryanCutler (approved):
Just a couple questions - I think I answered some myself as I went along :)
The code changes look good to me, I didn't look at the notebooks but I'm assuming they are updating the apis, so feel free to merge.

text_extensions_for_pandas/array/span.py
```diff
@@ -168,6 +206,10 @@ def end(self):
     def target_text(self):
         return self._text
 
+    @property
+    def document_text(self):
```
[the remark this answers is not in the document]
frreiss: Actually, I'm not sure why I added that property to `Span`. Removing it.

text_extensions_for_pandas/array/span.py
frreiss: Good idea. Shortened further to
```python
                return all(elem is not None and not isinstance(elem, Span)
                           for elem in seq)
```

text_extensions_for_pandas/array/span.py
frreiss: Fixed.

text_extensions_for_pandas/array/span.py
frreiss: Yes, I suppose returning `None` for the multi-document case could lead to trouble. I'll change this property so that it raises an exception if there isn't exactly one document. I'll also change `TokenSpanArray.document_tokens` to match. The purpose of this property is to provide a shorthand for the common pattern, "check that every span in this array is over the same document, then grab that document's text".

You can get a list of all the distinct document texts with:
``` python
(s.document_text for s in my_array.split_by_document())
```
though it might be a good idea to add a dedicated "enumerate just the distinct document texts" method at some point.

text_extensions_for_pandas/array/span.py
frreiss: The `@memoized_property` decorator will add the field back the next time the property is accessed. See https://github.com/estebistec/python-memoized-property/blob/17424f077037a361cd09f03e4e6376e76361281c/memoized_property.py#L7

Yes, it's not pretty. For what it's worth, the new `@functools.cached_property` decorator in Python 3.8 is worse. It stores the cached value in a field with the same name as the property.

frreiss:
Thanks for the careful review, @BryanCutler ! I'm going to merge these changes into master now.

PR #171, opened 2021-01-28
Fixes for use with Pandas 1.2.1

### LINKED ISSUE #168: Look into test failures with Pandas 1.2.1

BryanCutler:
Fixed `_from_sequence` to handle np.nan, added `__contains__` to check for null values with text set to `None`.

Closes #168

### DIFF
--- a/text_extensions_for_pandas/array/span.py
+++ b/text_extensions_for_pandas/array/span.py
@@ -466,6 +466,17 @@ class SpanArray(pd.api.extensions.ExtensionArray, SpanOpMixin):
                                self._ends.tobytes()))
         return self._hash
 
+    def __contains__(self, item) -> bool:
+        """
+        Return true if scalar item exists in this SpanArray.
+        :param item: scalar Span value.
+        :return: true if item exists in this SpanArray.
+        """
+        if isinstance(item, Span) and \
+                item.begin == Span.NULL_OFFSET_VALUE:
+            return Span.NULL_OFFSET_VALUE in self._begins
+        return super().__contains__(item)
+
     def equals(self, other: "SpanArray"):
         """
         :param other: A second `SpanArray`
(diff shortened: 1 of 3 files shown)


BryanCutler:
@frreiss there is still one more error, it has to do with TensorArray repr() with floating point values, and is similar to https://github.com/pandas-dev/pandas/pull/38391. I need to get back on that for a fix, but I'll try for a workaround first. We can merge this first if ok with you.

text_extensions_for_pandas/array/span.py
```diff
@@ -466,6 +466,17 @@ def __hash__(self):
...
 
+    def __contains__(self, item) -> bool:
+        """
+        Return true if scalar item exists in this SpanArray.
+        :param item: scalar Span value.
+        :return: true if item exists in this SpanArray.
+        """
+        if isinstance(item, Span) and \
+                item.begin == Span.NULL_OFFSET_VALUE:
+            return Span.NULL_OFFSET_VALUE in self._begins
```
BryanCutler, line 477: This is needed because otherwise the default is to check all values for equality, and that will fail if there is a NULL value with `Span.target_text=None`

text_extensions_for_pandas/array/token_span.py
```diff
@@ -483,14 +494,18 @@ def _from_sequence(cls, scalars, dtype=None, copy=False):
...
+                # TODO: Temporary fix for np.nan values, pandas-dev GH#38980
+                if np.isnan(s):
+                    s = TokenSpanDtype().na_value
+                else:
+                    raise ValueError(
+                        f"Can only convert a sequence of TokenSpan "
+                        f"objects to a TokenSpanArray. Found an "
+                        f"object of type {type(s)}"
+                    )
+            if tokens is None and not (s.begin == TokenSpan.NULL_OFFSET_VALUE and s.tokens.target_text is None):
```
BryanCutler, line 506: This gets a little awkward dealing with with a NULL value with target_text=None, but should be ok for now

text_extensions_for_pandas/array/span.py
frreiss: Looks like the current Pandas policy is "¯\_(ツ)_/¯"; see the warning on https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html

BryanCutler: Yeah, good question. The test failure related to this was checking `assert na_value in data_missing` which expects that `na_value == na_value` to be True. Seems like it would be different for `np.nan` though..

BryanCutler:
Going to go ahead and merge, will try to fix the TensorArray issue separately

PR #172, opened 2021-01-29
Patch Pandas ExtensionArrayFormatter to handle ndims > 1

### LINKED ISSUE #151: 2+ dimensional tensors of timestamps crash pd.Series.repr()

BryanCutler:
Pandas formatting has been failing in TensorArray for datetime64 values and float values with ndims > 1. This patches the `ExtensionArrayFormatter` to handle higher dimension formatting until a proper fix exists upstream.

Closes #151


BryanCutler:
@frreiss I'm not crazy about patching like this, but since we can't display TensorArrays with float values of ndim > 1, that is a big limitation. I'm still working on the upstream fix, but waiting on that might prevent us from upgrading Pandas for a while. wdyt?

BryanCutler:
Still some kinks to work out with text spacing in the tests, I think..

BryanCutler:
I think I got all the kinks worked out. I had to add some special handling for pandas 1.0.x, but that can all be removed once we bump the minimum version.

text_extensions_for_pandas/array/tensor.py
```diff
@@ -29,6 +30,115 @@
...
+            tpl = "{v}" if leading_space is False else " {v}"
+            fmt_values.append(tpl.format(v=_format(v)))
+        return fmt_values
+
+    return format_strings_slim(fmt_array, self.leading_space)
+
+
+ExtensionArrayFormatter._format_strings_orig = \
+    ExtensionArrayFormatter._format_strings
+if LooseVersion(pd.__version__) >= LooseVersion("1.1.0"):
```
frreiss, line 135: I think it would be safer for this code to disable the patch for versions of Pandas older than 1.0 and versions newer than 1.2.1 (i.e. versions that haven't been released yet).

frreiss (approved):
LGTM. You're right that this sort of thing is to be avoided if possible, but I don't see how we an avoid this and still have compatibility with Pandas versions between 1.0 and 1.2.1.

The main risk is that the patched function might break some other code in the user's environment that doesn't use Text Extensions for Pandas at all. That type of crash would be a really nasty one to track down the root cause of.

Have you run the full regression suite of Pandas itself with this patch in place?

text_extensions_for_pandas/array/tensor.py
BryanCutler: Good point about possibly messing up other user code, I can also add a check to see if the values are coming from the TensorArray.

Adding more strict version checking for future might be good too, but since we are not sure when it will be fixed, it might be good to leave the upper bound open? I could also add an env var to not patch as a fallback for the user.

BryanCutler: > Have you run the full regression suite of Pandas itself with this patch in place?

No, but that's a good idea too, I'll do that

frreiss: I like the environment variable idea.

BryanCutler:
@frreiss I added a check to use the original method if values aren't from a TensorArray, added the env var to switch it off completely, and added an upper bound to stop patching at Pandas 1.3,0+. I'm hopeful I'll be able to get a fix in by then :)  If not we can just bump that version up. I'm going to go ahead and merge this and we can start using newer pandas versions.

text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -224,9 +223,9 @@ def test_repr(self):
         result = repr(pd.Series(s))
         expected = textwrap.dedent(
             """\
-            0   [1 2]
-            1   [3 4]
-            2   [5 6]
+            0    [1, 2]
```
BryanCutler, line 226: the leading space is because that option defaults to true, we might want to look into changing that for tensors at least

PR #175, opened 2021-02-12
Remove Pandas upper version limit, bump testing versions

BryanCutler:


BryanCutler:
@frreiss now that #172 is merged, this removes the Pandas upper version cap in requirements, and bumps CI testing to the latest 2 minor versions of Pandas. This is so we can begin testing against the most recent versions.

frreiss:
LGTM.

PR #178, opened 2021-02-19
Fixed bug when adding spans from a series

BryanCutler:
In the addition operation, the span should return NotImplemented to allow Pandas to unwrap

### DIFF
--- a/text_extensions_for_pandas/array/test_span.py
+++ b/text_extensions_for_pandas/array/test_span.py
@@ -387,6 +387,50 @@ class CharSpanArrayTest(ArrayTestBase):
             one_one_2_2.contains(Span(test_text, 1, 1)), [True, False]
         )
 
+    def test_addition(self):
+        arr = self._make_spans_of_tokens()
+
+        # Test SpanArray + SpanArray
+        a = arr[:2]
+        b = arr[2:]
+        result = a + b  # ["This is", "a test"]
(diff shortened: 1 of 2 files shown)


text_extensions_for_pandas/array/test_span.py
```diff
@@ -387,6 +387,50 @@ def test_contains(self):
             one_one_2_2.contains(Span(test_text, 1, 1)), [True, False]
         )
 
+    def test_addition(self):
+        arr = self._make_spans_of_tokens()
+
+        # Test SpanArray + SpanArray
+        a = arr[:2]
+        b = arr[2:]
+        result = a + b  # ["This is", "a test"]
```
BryanCutler, line 396: Is this right?

frreiss:
```suggestion
        result = a + b  # [[0, 9) 'This is a', [5, 14) 'is a test']
```
`a` is `[ [0, 4) 'This', [5, 7) 'is' ]` and `b` is `[ [8, 9) 'a', [10, 14) 'test']`.

The assertions 2 and 3 lines down are correct.

frreiss (approved):
LGTM.

text_extensions_for_pandas/array/test_span.py
BryanCutler: Ah, just my comment was wrong, doh

PR #180, opened 2021-03-10
Improvements to dictionary matching

frreiss:
This PR includes some fixes for issues with our dictionary matching that encountered while working on a market intelligence use case for a blog post.

The specific problems addressed here are:
* There was no API for creating a dictionary from an in-memory list, so I've added a new function `create_dict()`.
* SpaCy's default set tokenizers tend to overthink how they treat punctuation, which leads to dictionary entries getting tokenized differently from text, which messes up dictionary matching. So I've added a function `simple_tokenizer()` that returns a tokenizer that splits on every chunk of whitespace and on every punctuation character. Dictionary creation uses that tokenizer by default now.

I also fixed a minor bug in the handling of the `warnings` element of responses from Watson Natural Language Understanding.


frreiss:
@BryanCutler would you mind giving these changes a quick review?

Thanks!

BryanCutler:
Sorry, slipped under the radar. Looking at it now.

text_extensions_for_pandas/spanner/extract.py
```diff
@@ -52,17 +53,42 @@ def load_dict(file_name: str, tokenizer: "spacy.tokenizer.Tokenizer"):
...
+    English tokenizer.
 
-    :return: a `pd.DataFrame` with the normalized entries.
+    :return: a `pd.DataFrame` with the normalized, tokenized dictionary entries.
     """
     with open(file_name, "r") as f:
         lines = [
-            line.strip() for line in f.readlines() if len(line) > 0 and line[0] != "#"
+            line.strip() for line in f.readlines()
+            if len(line.strip()) > 0 and line[0] != "#"
```
BryanCutler, line 64: nit: should this be indented?

text_extensions_for_pandas/spanner/extract.py
```diff
@@ -52,17 +53,42 @@ def load_dict(file_name: str, tokenizer: "spacy.tokenizer.Tokenizer"):
...
+    """
+    Create a dictionary from a list of entries, where each entry is expressed as a
+    single string.
+
+    Tokenizes and normalizes the dictionary entries.
+
+    :param entries: Iterable of strings, one string per dictionary entry.
+
+    :param tokenizer: Preconfigured tokenizer object for tokenizing
+    dictionary entries.  **Must always tokenizer the same way as the tokenizer
```
BryanCutler, line 81: i think the grammer is off. Do you mean "Must always tokenize the same way as the tokenizer..."?

BryanCutler (approved):
Just a couple nit questions, otherwise LGTM

text_extensions_for_pandas/spanner/extract.py
frreiss: Reformatted this file with `black`.

text_extensions_for_pandas/spanner/extract.py
frreiss: Oops! Fixed now.

frreiss:
Thanks for the review! Pushed some corrections. Will merge once this branch passes tests.

Issue #179, opened 2021-03-03
Reimplement Arrow conversion to cover multi-document SpanArrays/TokenSpanArrays


frreiss:
This item covers the remaining tasks from #73:
* Extend the code in `arrow_conversion.py` to cover multi-document arrays
* Re-enable the tests of feather I/O: `CharSpanArrayIOTests.test_feather()` and `TokenSpanArrayIOTests`

PR #181, opened 2021-03-23
Fix Arrow serializaiton for SpanArray multidoc support

BryanCutler:
This changes Arrow serialization for SpanArray to store documents in a dictionary that is indexed by text ids. Also added support for saving to Parquet files.

From #179

### DIFF
--- a/text_extensions_for_pandas/array/arrow_conversion.py
+++ b/text_extensions_for_pandas/array/arrow_conversion.py
@@ -158,6 +158,14 @@ def arrow_to_span(extension_array: pa.ExtensionArray) -> SpanArray:
             raise ValueError("Only pyarrow.Array with a single chunk is supported")
         extension_array = extension_array.chunk(0)
 
+    # NOTE: workaround for bug in parquet reading
+    if pa.types.is_struct(extension_array.type):
+        index_dtype = extension_array.field(ArrowSpanType.BEGINS_NAME).type
+        target_text_dict_dtype = extension_array.field(ArrowSpanType.TARGET_TEXT_DICT_NAME).type
+        extension_array = pa.ExtensionArray.from_storage(
+            ArrowSpanType(index_dtype, target_text_dict_dtype),
+            extension_array)
+
     assert pa.types.is_struct(extension_array.storage.type)
 
     # Get the target text dictionary array
--- a/text_extensions_for_pandas/array/arrow_conversion.py
+++ b/text_extensions_for_pandas/array/arrow_conversion.py
@@ -27,41 +27,40 @@ import pyarrow as pa
 from text_extensions_for_pandas.array.span import SpanArray
 from text_extensions_for_pandas.array.token_span import TokenSpanArray
 from text_extensions_for_pandas.array.tensor import TensorArray
+from text_extensions_for_pandas.array.string_table import StringTable
 
 
 class ArrowSpanType(pa.PyExtensionType):
     """
     PyArrow extension type definition for conversions to/from Span columns
     """
+    BEGINS_NAME = "span_begins"
+    ENDS_NAME = "span_ends"
+    TARGET_TEXT_DICT_NAME = "target_text"
 
-    TARGET_TEXT_KEY = b"target_text"  # metadata key/value gets serialized to bytes
-    BEGINS_NAME = "char_begins"
-    ENDS_NAME = "char_ends"
-
-    def __init__(self, index_dtype, target_text):
+    def __init__(self, index_dtype, target_text_dict_dtype):
         """
         Create an instance of a Span data type with given index type and
-        target text that will be stored in Field metadata.
+        target text dictionary type. The dictionary type will hold target text ids
+        that map to a dictionary of document target texts.
 
-        :param index_dtype:
-        :param target_text:
+        :param index_dtype: type for the begin, end index arrays
+        :param target_text_dict_dtype: type for the target text dictionary array
         """
         assert pa.types.is_integer(index_dtype)
 
-        # Store target text as field metadata
-        metadata = {self.TARGET_TEXT_KEY: target_text}
-
         fields = [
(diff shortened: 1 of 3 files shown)


text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -133,10 +132,15 @@ def span_to_arrow(char_span: SpanArray) -> pa.ExtensionArray:
     begins_array = pa.array(char_span.begin)
     ends_array = pa.array(char_span.end)
 
-    typ = ArrowSpanType(begins_array.type, char_span.target_text)
+    # Create a dictionary array from StringTable used in this span
+    dictionary = pa.array(list(char_span._string_table.things))
+    target_text_dict_array = pa.DictionaryArray.from_arrays(char_span._text_ids, dictionary)
+    # TODO: remove unused things and normalize text_ids?
```
BryanCutler, line 138: Not sure if we should sanitize these before serializing to remove any unused document, then reset the text ids?

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -133,10 +132,15 @@ def span_to_arrow(char_span: SpanArray) -> pa.ExtensionArray:
     begins_array = pa.array(char_span.begin)
     ends_array = pa.array(char_span.end)
 
-    typ = ArrowSpanType(begins_array.type, char_span.target_text)
+    # Create a dictionary array from StringTable used in this span
+    dictionary = pa.array(list(char_span._string_table.things))
+    target_text_dict_array = pa.DictionaryArray.from_arrays(char_span._text_ids, dictionary)
```
BryanCutler, line 137: This uses the private `char_span._string_table` and `._text_ids`. It didn't seem like we really needed accessors for these so I left it like this.

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -154,13 +158,21 @@ def arrow_to_span(extension_array: pa.ExtensionArray) -> SpanArray:
             raise ValueError("Only pyarrow.Array with a single chunk is supported")
         extension_array = extension_array.chunk(0)
 
+    # NOTE: workaround for bug in parquet reading
+    if pa.types.is_struct(extension_array.type):
+        index_dtype = extension_array.field(ArrowSpanType.BEGINS_NAME).type
+        target_text_dict_dtype = extension_array.field(ArrowSpanType.TARGET_TEXT_DICT_NAME).type
+        extension_array = pa.ExtensionArray.from_storage(
+            ArrowSpanType(index_dtype, target_text_dict_dtype),
+            extension_array)
```
BryanCutler, line 167: For some reason, parquet doesn't return an extension array, but the struct array used as storage. Seems like a bug, so I'll follow up on it, but this workaround seemed ok for now.

text_extensions_for_pandas/array/arrow_conversion.py
```diff
@@ -154,13 +158,21 @@ def arrow_to_span(extension_array: pa.ExtensionArray) -> SpanArray:
...
 
-    # Get target text from the begins field metadata and decode string
-    metadata = extension_array.storage.type[ArrowSpanType.BEGINS_NAME].metadata
-    target_text = metadata[ArrowSpanType.TARGET_TEXT_KEY]
-    if isinstance(target_text, bytes):
-        target_text = target_text.decode()
+    # Create target text StringTable and text_ids from dictionary array
+    target_text_dict_array = extension_array.storage.field(ArrowSpanType.TARGET_TEXT_DICT_NAME)
+    target_texts = [s.as_py() for s in target_text_dict_array.dictionary]
+    string_table, _ = StringTable.merge_things(target_texts)
```
BryanCutler, line 174: This assumes that the `text_ids` will match the new ids returned from `merge_thing`. I think it's safe to do this.

BryanCutler:
@frreiss this seems like a good improvement for SpanArray serialization - much better to store in a dictionary batch rather than field metadata. If this looks ok, I'll get started on TokenSpanArray.

text_extensions_for_pandas/array/test_span.py
```diff
@@ -455,17 +457,28 @@ def test_addition(self):
 
 class CharSpanArrayIOTests(ArrayTestBase):
 
-    @pytest.mark.skip("Temporarily disabled until Feather support reimplemented")
     def test_feather(self):
         arr = self._make_spans_of_tokens()
```
frreiss, line 461: Would you mind modifying these tests so that they write out a `SpanArray` containing spans over two different document texts?

frreiss (approved):
Looking good. Some minor cleanup requested.

text_extensions_for_pandas/array/arrow_conversion.py
BryanCutler: yup, that's a good idea. I'll add that here.

text_extensions_for_pandas/array/arrow_conversion.py
BryanCutler: Well the dictionary in pyarrow is just an ordinary array, we just need to make sure that the indices match those in the `_text_ids` array, which I think they do. I believe I should unbox the "thing" first though.

text_extensions_for_pandas/array/arrow_conversion.py
BryanCutler: It's a little more useful to have a method to create a table from just a list of strings, so I added that. If you think we should have just one that takes things and ids to make a table, I can change it.

text_extensions_for_pandas/array/test_span.py
BryanCutler: I added a separate test for multi-doc.

BryanCutler:
I think I addressed all and tests are passing. I'll go ahead and merge now and fix up with a followup or when I fix TokenSpanArray arrow conversion.

PR #182, opened 2021-03-30
[WIP] Fix Arrow conversion for TokenSpanArray with multi-doc

BryanCutler:
This fixes Arrow conversion of a TokenSpanArray that possibly uses multiple SpanArrays with multiple documents. The conversion follows these steps:

1. Get SpanArrays used in the TokenSpanArray and convert these to arrow, each is a struct with begins, ends, and dictionary array with the document
2. Concat the arrow SpanArrays together to make an arrow list of them
3. Build an arrow dictionary array that maps the index of the SpanArray used by each element of the TokenSpanArray to the SpanArray in the arrow list
4. create a struct array with token begins, token ends, and the token dictionary array

### DIFF
--- a/text_extensions_for_pandas/array/arrow_conversion.py
+++ b/text_extensions_for_pandas/array/arrow_conversion.py
@@ -70,6 +70,40 @@ class ArrowTokenSpanType(pa.PyExtensionType):
 
     BEGINS_NAME = "token_begins"
     ENDS_NAME = "token_ends"
+
+    def __init__(self, index_dtype, token_dict_dtype):
+        """
+        Create an instance of a TokenSpan data type with given index type and
+        target text that will be stored in Field metadata.
+
+        :param index_dtype:
+        :param target_text:
+        """
+        assert pa.types.is_integer(index_dtype)
+
+        token_span_fields = [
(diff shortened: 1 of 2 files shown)


BryanCutler:
@frreiss this is still a bit of a WIP, need to add some tests like actual multi-doc array. Please have a look at the implementation when you can. It's a little complicated since we have dictionary arrays that point to other dictionary arrays, but overall I think it's a better approach than before.

Some areas are rough due to some issues in pyarrow, that probably could be made into a JIRA to be improved upstream, but I'll have to look into it more. Also, it seems this only works with pyarrow >= 2.0.0, so we might want to bump up the minimum version.

frreiss (approved):
Looks good. I didn't know you could stuff structured data into Arrow dictionaries -- saves a lot of work :)

frreiss:
Merging this in its current form, as we need to have at least a semi-functional version of serialization in master so we can cut another release. @BryanCutler can you make a second PR with those testing improvements when they're ready?

BryanCutler:
Thanks @frreiss . Yeah, I'll do a followup with testing requirements. What do you think of bumping the minimum pyarrow to 2.0.0? Alternatively, we could just raise an error if the user tries to serialize with a lesser version.

PR #185, opened 2021-04-10
Add new market intelligence use case

frreiss:
This PR adds parts 1 and 2 of a market intelligence use case involving Text Extensions for Pandas, SpaCy, and Watson Natural Language Understanding.

I've also made adjustments to the scripts for setting up Python environments in support of this use case, and I've expanded the readme files for the tutorials.

Since we're planning some Ray usage for later parts of the market intelligence series, I've added Ray to the development environment dependencies and modified some of the notebooks in the `tutorials/corpus` directory to use Ray for faster model training.


frreiss:
Looks like the SpaCy upgrade broke some tests. Pushed changes that should fix them.

text_extensions_for_pandas/io/test_conll.py
```diff
@@ -50,7 +50,7 @@ def test_iob_to_spans(self):
                 """\
                                           span ent_type
                 0           [61, 67): 'Alaska'      GPE
-                1      [73, 84): 'Santa Claus'   PERSON
+                1      [73, 84): 'Santa Claus'      ORG
```
BryanCutler, line 53: I suppose this is more accurate, haha!

BryanCutler (commented):
Code changes look good. Will review the notebooks now.

text_extensions_for_pandas/io/test_conll.py
frreiss: :) Updates to the `en_core_web_sm` model tend to be "two steps forward, one step back".

BryanCutler (approved):
Did a quick pass on reviewNB, and notebooks look good. I'll try to run through market use case 2 once you merge these.

PR #187, opened 2021-04-14
Adding NumPy ufunc to TensorArray

BryanCutler:
This implements the `__array_ufunc` method in `TensorArray` for better integration with NumPy. When TensorArrays are used with numpy universal functions, such as `np.add()`, this method will be invoked to unwrap the inputs then wrap back the output as a TensorArray.

### DIFF
--- a/text_extensions_for_pandas/array/test_tensor.py
+++ b/text_extensions_for_pandas/array/test_tensor.py
@@ -375,7 +375,10 @@ class TestTensor(unittest.TestCase):
 
         # Test as agg to TensorElement, defaults to axis=0
         result = s % 2 == 0
-        npt.assert_array_equal(result.all(), np.array([True, False]))
+        a = np.asarray(arr)
+        r = np.all(result.array, axis=1)
+        r = np.max(arr, axis=1)
+        #npt.assert_array_equal(result.all(), np.array([True, False]))
 
     def test_any(self):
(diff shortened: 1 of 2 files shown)


BryanCutler:
@frreiss this is to improve numpy integration with TensorArray. I might touch on this in my blog so will be good to add this.

text_extensions_for_pandas/array/test_tensor.py
```diff
@@ -499,6 +499,28 @@ def test_inferred_type(self):
         arr = TensorArray([True, False, True])
         self.assertEqual(arr.inferred_type, "boolean")
 
+    def test_numpy_ufunc(self):
+        data = np.arange(10).reshape(5, 2)
+        arr = TensorArray(data)
+
+        # ufunc with number
```
frreiss, line 506: It would be a good idea to repeat these tests with the arguments in the opposite order, i.e. `np.add(1, arr)`.

frreiss (approved):
LGTM, thanks for adding this!

text_extensions_for_pandas/array/test_tensor.py
BryanCutler: I didn't think it was necessary since the inputs are handled as a list, but I can add them - it wouldn't hurt

PR #192, opened 2021-05-06
Expand supported Python versions to cover 3.6 and increment package version

frreiss:
While doing some internal testing, I discovered that we didn't update the Python version in `setup.py` when we added Python 3.6 support. This PR updates the package metadata to indicate that this library runs on Python 3.6.


frreiss:
I'm going to merge this change now so I can cut another beta release and perform downstream testing with Python 3.6.

BryanCutler (commented):
Good catch, LGTM!

Issue #191, opened 2021-04-26
Update CoNLL-U support to fully cover EWT data set


frreiss:
`text_extensions_for_pandas.io.conll.conll_2003_to_dataframes` currently has trouble parsing data files from the EWT data set for shallow semantic parsing (https://github.com/UniversalDependencies/UD_English-EWT).

We need to update this support to cover this data set.

Recommended approach:
* Create a new entry point specific to the format used in EWT -- say, `conll_u_to_dataframes()`
* Add a new argument to cover the document metadata that EWT includes in comments
* Add support for optional fields at the end of the record, repeated any number of times in sequence
* Factor out common code with the existing `conll_2003_to_dataframes` into a shared internal entry point

The change set for this item should include a test case that downloads and parses part of the EWT data set when run. Be sure to cache the downloaded files. Be sure to avoid checking in EWT data to the Text Extensions for Pandas repository.

frreiss:
Some specific data sets that should work out of the box:
* [Universal Dependencies treebank](https://universaldependencies.org/#download)
* [English Web Treebank (EWT)](https://github.com/UniversalDependencies/UD_English-EWT)
* [CoNLL-09](https://ufal.mff.cuni.cz/conll2009-st/)
* [OntoNotes](https://catalog.ldc.upenn.edu/LDC2013T19) (this one requires a license to access)

frreiss:
Known "gotchas" with CoNLL-U:
File encoding should be UTF-8, but can't rely on that.

Comments often contain data. For example, the first document in the dev fold of EWT:
```
# newdoc id = weblog-blogspot.com_nominations_20041117172713_ENG_20041117_172713
# sent_id = weblog-blogspot.com_nominations_20041117172713_ENG_20041117_172713-0001
# newpar id = weblog-blogspot.com_nominations_20041117172713_ENG_20041117_172713-p0001
# text = From the AP comes this story :
1	From	from	ADP	IN	_	3	case	3:case	_
2	the	the	DET	DT	Definite=Def|PronType=Art	3	det	3:det	_
3	AP	AP	PROPN	NNP	Number=Sing	4	obl	4:obl:from	_
4	comes	come	VERB	VBZ	Mood=Ind|Number=Sing|Person=3|Tense=Pres|VerbForm=Fin	0	root	0:root	_
5	this	this	DET	DT	Number=Sing|PronType=Dem	6	det	6:det	_
[...]

PR #194, opened 2021-06-01
Update Conll-U to fully support and cover EWT dataset

ZachEichen:
Addresses issue #191 and adds support for importing CoNLL-U data-format files, especially those in EWT, Global Dependencies, and conll_2009 formats, as well as Ontonotes. 

Created a separate method from the conll_to_df  as a new entry-point for these dataformats, which supports similar options to other available packages supporting .conllu files (such as Spacy). Refactored common code within the io/conll module to separate methods.


text_extensions_for_pandas/io/conll.py
```diff
@@ -79,19 +86,28 @@ class _SentenceData:
...
-        self._token_metadata = _make_empty_meta_values(column_names, iob_columns)
-
         # Line numbers for each token from the file
         self._line_nums = []  # Type: List[int]
 
+        # metadata from conll_u file
+        self._sentence_id = None
+        self._paragraph_id = None
+        self._doc_id = None
+        self.conll_09_format = predicate_args
```
frreiss, line 108: This class field name should start with underscore.

text_extensions_for_pandas/io/conll.py
```diff
@@ -108,19 +124,35 @@ def token_metadata(self) -> Dict[str, List[str]]:
...
+
+    @property
+    def column_names(self):
+        return self._column_names
+
+    @property
+    def has_conll_u_metadata(self):
+        return (self._sentence_id is not None) or (self._paragraph_id is not None) or (self._doc_id is not None)
+
+    def set_conll_u_metadata(self, doc_id: str = None, paragraph_id: str = None, sent_id: str = None):
```
frreiss, line 147: Would you mind augmenting these fields with a configurable set of key-value pairs? The CoNLL-U spec seems to think the comments before each sentence are supposed to hold arbitrary named attributes; see https://universaldependencies.org/format.html#sentence-boundaries-and-comments

text_extensions_for_pandas/io/conll.py
```diff
@@ -148,6 +180,58 @@ def add_line(self, line_num: int, line_elems: List[str]):
...
+                             f"{1 + len(self._column_names)} elements.")
+        token = line_elems[0]
+        raw_tags = line_elems[1:]
+        self._tokens.append(token)
+        self._line_nums.append(line_num)
+        self._process_line_tags(raw_tags, line_num, line_elems, is_ewt=False)
+
+    def add_line_ewt(self, line_num: int, line_elems: List[str]):
+        """
+        :param line_num: Location in file, for error reporting
```
frreiss, line 200: Can you add a note here to tell how this is different from `add_line`?

text_extensions_for_pandas/io/conll.py
```diff
@@ -148,6 +180,58 @@ def add_line(self, line_num: int, line_elems: List[str]):
...
+        :param line_elems: Fields of a line, pre-split
+        """
+        if len(line_elems) < 2 + len(self._column_names):
+            if len(line_elems) > 2 + self._num_standard_cols:
+                line_elems.extend(['_' for i in range(2 + len(self._column_names) - len(line_elems))])
+                print(f"Unexpected number of elements {len(line_elems)} "
```
frreiss, line 206: It doesn't look like this `print` statement is supposed to be here.

text_extensions_for_pandas/io/conll.py
```diff
@@ -217,6 +301,122 @@ def _parse_conll_file(input_file: str,
...
+def _parse_conll_u_file(input_file: str,
+                        column_names: List[str],
+                        iob_columns: List[bool],
+                        predicate_args:bool = True,
+                        merge_subtokens: bool = False,
+                        merge_subtoken_seperator: str = '|') \
+        -> List[List[_SentenceData]]:
+    """
+
+    Parses EWT file format to python objects
```
frreiss, line 313: Is this function just for EWT, or is it for anything that meets the CoNLL-U standard? If the former, the function ought to be called `_parse_ewt_file`; if the latter, this docstring needs to be updated.

text_extensions_for_pandas/io/conll.py
```diff
@@ -40,6 +40,7 @@
 # Special token that CoNLL-2003 format uses to delineate the documents in
 # the collection.
 _CONLL_DOC_SEPARATOR = "-DOCSTART-"
+_EWT_DOC_SEPERATOR = "# newdoc id"
```
frreiss, line 43: I think this separator is actually officially part of the CoNLL-U standard and isn't just for EWT; see https://universaldependencies.org/format.html#paragraph-and-document-boundaries

text_extensions_for_pandas/io/conll.py
```diff
@@ -217,6 +301,122 @@ def _parse_conll_file(input_file: str,
...
+                doc_id = line_elems[1]
+                current_sentence.set_conll_u_metadata(doc_id=doc_id)
+                paragraph_id = ''
+            elif line_elems[0] == "# newpar id":
+                paragraph_id = line_elems[1]
+                current_sentence.set_conll_u_metadata(paragraph_id=paragraph_id)
+            elif line_elems[0] == "# sent_id":
+                sentence_id = line_elems[1]
+                current_sentence.set_conll_u_metadata(sent_id=sentence_id)
+
```
frreiss, line 381: There should be an additional branch to this `if` statement that gathers up any other metadata key-value pairs that the dataset sees fit to attach to the sentence (see https://universaldependencies.org/format.html#sentence-boundaries-and-comments).

It would be nice if there was also a way to have this function map a user-configurable set of additional metadata values to additional columns of the returned DataFrame.

text_extensions_for_pandas/io/conll.py
```diff
@@ -724,6 +960,66 @@ def conll_2003_to_dataframes(input_file: str,
...
+     the returned dataframe will contain *two* columns, holding **IOB2** tags and
+     entity type tags, respectively. For example, an input column "ent" will turn into
+     output columns "ent_iob" and "ent_type". By default in CONLL_U or EWT formats this is all false.
+    :param has_predicate_args: Whether or not the file format includes predicate args. True by default, and
+     should support most files in the conllu format, but will assume that any tabs in the last
+     element are additional predicate arguments
+    :param merge_subtokens: dictates how to handle tokens that are smaller than one word. By default, we keep
+     the subtokens as two seperate entities, but if this is set to true, the subtokens will be merged into a
+     single entity, of the same length as the token, and their attributes will be concatenated
+    :param merge_subtoken_seperator: If merge subtokens is selected, concatenate the attributes with this
```
frreiss, line 993: seperator ==> separator

frreiss (approved):
Looking good. Some minor changes requested inline.

Would you mind also running the modified files through `black` before checking in?

Issue #195, opened 2021-06-02
Create an example notebook that shows training a classifier over the EWT dataset


frreiss:
Create a notebook that is similar to `tutorials/corpus/CoNLL_3.ipynb`, with the following differences:
* The new notebook uses the semantic role labeling version of the EWT data set as its input
* The new notebook trains a token classifier over the `upostag` column

The new notebook should share the code in `tutorials/corpus/util.py`. Modify that shared code as needed.

frreiss:
Closed by #198

PR #198, opened 2021-06-03
Create demo notebook for Conllu files

ZachEichen:
Created a demo notebook that shows how the new conllu logic can be used to open and process diverse .conllu files; addresses issue #195 

Shows how these can be powerfully used with pandas dataframes, and other text-extensions integrations to quickly manipulate the data and do things like displaying parse trees, and retokenizing with spaCy.


notebooks/Read_CoNLL-U_files.ipynb
```diff
@@ -0,0 +1,1116 @@
+{
```
frreiss, line 1: <p>This approach will work for now, but it's looking like we should have a proper "datasets" package to reduce the amount of duplicate code we're putting into all these notebooks. Marked that work down for later in https://github.com/CODAIT/text-extensions-for-pandas/issues/199</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

frreiss, line 1: <p>It looks like the "head" columns ended up as floating point values somehow. They should be ints. Did some nulls sneak into the head pointers?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

frreiss, line 1: <p>Can you add "document number" and "fold name" columns here so that these DataFrames can be separated back out later?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

frreiss, line 1: <p>Maybe add a <code>%%time</code>  magic to this cell so that the speed of Parquet reading comes across in the static render of the notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CODAIT/text-extensions-for-pandas/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ZachEichen: I believe this actually has to do with 'virtual' tokens where words were omitted from the surface representation and had to be added by the labeller, which have numbers of  the form 10.1 or similar

ZachEichen: Do you have a suggestion of how I should handle this?

ZachEichen: <p>For the first part of the notebook demo, I'm only using the dev fold, so this shouldn't be an issue. Later on when I'm using multiple folds, I do exactly that </p><p> </p>

ZachEichen: <p>Done. </p>

frreiss: The real root cause is upstream in Pandas. There is old code in various parts of Pandas that casts ints to floats for operations that *might* produce NA values. Pandas should be converting to nullable int types (see https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html) instead. I've been meaning to submit PRs to fix those problems but haven't had time.

For now, the best fix at the application level is to convert the series that Pandas has "floatified" back to an integer dtype. Use `int` if there aren't supposed to be NAs. Use `Int64` (with a capital "I") if there might be NAs.

frreiss: Follow-up: Looking some more at the code, I think the conversion to a nullable integer dtype should happen inside `conll_u_to_dataframes`.

frreiss:
This is very close to ready to merge. Some minor requests:
* I see some formatting issues in `conll.py`. Can you run that file through `black`?
* Need an explicit conversion back to an int dtype to get rid of the unintentional float outputs to `conll_u_to_dataframes` -- see https://github.com/CODAIT/text-extensions-for-pandas/pull/198#discussion_r647665495.
* Can you please add a precision/recall/F1 score computation after the last cell? Collection-level F1 is fine.

frreiss (approved):
LGTM.

PR #207, opened 2021-06-10
Add some missing extract ops

frreiss:
While working on a blog post, I noticed that we are missing some variants of the `extract` operator from spanner algebra:
* `extract_regex`: return all matches of a character-based regex
* `extract_split`: Split the document into spans according to a set of split points

This PR implements those two operations and adds appropriate tests.

### DIFF
--- a/text_extensions_for_pandas/spanner/extract.py
+++ b/text_extensions_for_pandas/spanner/extract.py
@@ -246,8 +246,8 @@ def extract_regex(
     compiled_regex: re.Pattern
 ):
     """
-    Identify all non-overlapping matches of a regular expression, as returned by `
-    `regex.finditer()``, and return those locations as an array of spans.
+    Identify all non-overlapping matches of a regular expression, as returned by
+    ``re.Pattern.finditer()``, and return those locations as an array of spans.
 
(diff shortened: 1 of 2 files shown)


text_extensions_for_pandas/spanner/extract.py
```diff
@@ -237,3 +239,76 @@ def extract_regex_tok(
...
+def extract_split(
+    doc_text: str, split_points: Union[Sequence[int], np.ndarray, SpanArray]
+) -> SpanArray:
+    """
+    Split a document into spans along a specified set of split points.
+
+    :param doc_text: Text of the document; will be the target text of the returned spans.
+
+    :param split_points: A series of offsets into ``doc_text``, expressed as either:
+      * A sequence if integers (split at certain locations and return a set of splits that
```
BryanCutler, line 277: should it be "sequence of integers"?

BryanCutler (commented):
LGTM, just noticed one possible typo. Looks cool!

text_extensions_for_pandas/spanner/extract.py
frreiss: Fixed.

BryanCutler (approved):

PR #213, opened 2021-06-21
Fixed multi-document spanarray rendering

PokkeFe:
Added a wrapper object that can handle rendering multi-document SpanArrays, and restructured the initialization script to match the new render method.


frreiss:
These changes look good and fix the crash `notebooks/Text_Extensions_for_Pandas_Overview.ipynb`.

I'd like to request two additional changes, though:
* Add a regression test that displays two documents in `test_jupyter.py` so that we don't see similar problems pop up in the future.
* Add a limit on the number of documents displayed, so that we don't crash the user's browser if they accidentally display the "span" column of an entire corpus.

frreiss (approved):
LGTM

PR #218, opened 2021-07-03
Fix failing tests with Pandas 1.3.0 (was: Add logic to deal with ABCIndexClass being renamed to ABCIndex)

frreiss:
Pandas 1.3.x renamed its abstract base class for indexes from `ABCIndexClass` to `ABCIndex`, which is messing up some of our type checks. This PR adds some logic to work around that renaming. I'm using exceptions as control flow, which is a bit ugly, but the alternatives are also ugly.

### DIFF
--- a/text_extensions_for_pandas/array/span.py
+++ b/text_extensions_for_pandas/array/span.py
@@ -30,7 +30,14 @@ import pandas as pd
 from memoized_property import memoized_property
 # noinspection PyProtectedMember
 from pandas.api.types import is_bool_dtype
-from pandas.core.dtypes.generic import ABCDataFrame, ABCIndexClass, ABCSeries
+from pandas.core.dtypes.generic import ABCDataFrame, ABCSeries
+try:
+    from pandas.core.dtypes.generic import ABCIndexClass
+except ImportError:
+    # ABCIndexClass changed to ABCIndex in Pandas 1.3
+    # noinspection PyUnresolvedReferences
+    from pandas.core.dtypes.generic import ABCIndex
+    ABCIndexClass = ABCIndex
 from pandas.core.indexers import check_array_indexer
 
 # Internal imports
--- a/text_extensions_for_pandas/array/span.py
+++ b/text_extensions_for_pandas/array/span.py
@@ -326,6 +326,7 @@ class SpanDtype(pd.api.extensions.ExtensionDtype):
         SpanArray.
         """
         from text_extensions_for_pandas.array.arrow_conversion import arrow_to_span
+
         return arrow_to_span(extension_array)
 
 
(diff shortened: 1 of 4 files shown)


frreiss:
Okay, that didn't work. Will try something else.

frreiss:
Fixed some bugs in `SpanArray` and `TokenSpanArray` that new regression tests in Pandas 1.3.0 brought to light. Now we're down to 62 failing tests.

frreiss:
Update: Most of the test failures seem to be due to a bug in Pandas 1.3.0. Due to some performance optimizations introduced in https://github.com/pandas-dev/pandas/pull/40353, Pandas turns `DataFrame.iloc[slice(x, y, z)]` into `__getitem__((..., slice(x, y, z))` on the ExtensionArray that backs any column defined with an extension type.

Code to reproduce:
```python
from pandas.api.extensions import ExtensionArray,ExtensionDtype

class MyExtensionDtype(ExtensionDtype):
    """Minimal extension dtype"""
    def __init__(self):
        pass
    
    @property
    def type(self):
        return int
    
    @property
    def name(self) -> str:
        return "MyExtensionDtype"
        
    @classmethod
    def construct_array_type(cls):
        return MyExtensionArray()

class MyExtensionArray(ExtensionArray, ExtensionScalarOpsMixin):
    """Minimal extension array that logs calls to __getitem__()"""
    @property
    def dtype(self):
        return MyExtensionDtype()
    
    def copy(self):
        return MyExtensionArray()
    
    def __len__(self):
        return 5
    
    def __getitem__(self, key):
        print(f"__getitem__ called with key '{key}'")
        return 42

arr = MyExtensionArray()
df = pd.DataFrame({"a": arr})
_ = df.iloc[:3]
```
which prints out:
```
__getitem__ called with key '(Ellipsis, slice(None, 3, None))'
```
It should print the following instead:
```
__getitem__ called with key 'slice(None, 3, None)'
```

I'll put in a workaround tomorrow and file a bug with Pandas.

FYI @BryanCutler @ZachEichen @PokkeFe @Crushellini

frreiss:
Update:
* Opened https://github.com/pandas-dev/pandas/issues/42430 to cover the slicing issue described above.
* Added a workaround for the Pandas issue to `TensorArray.__getitem__()`
* We're now down to 53 failing tests.

frreiss:
Update: Fixed another minor bug. Now we are down to 49 failing tests.

frreiss:
Update:
* Filed https://github.com/pandas-dev/pandas/issues/42437 and temporarily disabled the tests that that bug was causing to fail, since a workaround exists. See #221 for a workaround for that issue.
* Fixed a minor issue in `TensorArray.isna()` that was causing a number of regressions.

Now we're down to 2 failing tests.

frreiss:
All tests passing against Pandas 1.3.0 now. Merging this PR to unblock other PRs.

text_extensions_for_pandas/array/span.py
```diff
@@ -30,7 +30,14 @@
...
 from pandas.api.types import is_bool_dtype
-from pandas.core.dtypes.generic import ABCDataFrame, ABCIndexClass, ABCSeries
+from pandas.core.dtypes.generic import ABCDataFrame, ABCSeries
+try:
+    from pandas.core.dtypes.generic import ABCIndexClass
+except ImportError:
+    # ABCIndexClass changed to ABCIndex in Pandas 1.3
+    # noinspection PyUnresolvedReferences
+    from pandas.core.dtypes.generic import ABCIndex
+    ABCIndexClass = ABCIndex
```
BryanCutler, line 40: You could do `from pandas.core.dtypes.generic import ABCIndex as ABCIndexClass`

BryanCutler (commented):
LGTM, just a minor suggestion

text_extensions_for_pandas/array/span.py
frreiss: Oh, I should have thought of that. Will put in a follow-on PR with that edit.

PR #223, opened 2021-07-08
Fix regression from PR #218

frreiss:
This follow-on PR fixes a minor problem in the fix for `TokenSpanArray.is_single_document` from #218.


BryanCutler (approved):
LGTM

PR #224, opened 2021-07-08
Handling for newlines in doctext + JS replace improvement

PokkeFe:
Added handling for newlines in SpanArray doctext rendering and updated the frontend's HTML sanitization with global regular expressions.


frreiss (approved):
These changes make sense to me. I'm going to merge these changes now to unblock the release, but there are two pieces of follow-on work to do here:
* Check in updated versions of notebooks that are affected by this change
* Update our regression tests (either add a new test or tweak the existing ones) so they will catch this bug if it arises again

PR #208, opened 2021-06-15
Multiple minor fixes to notebooks under tutorials directory

frreiss:
This PR groups together a number of minor fixes to the notebooks and Python files under the `tutorials` directory. Summary of changes:
* Fixed path issue that caused `CoNLL_2.ipynb` not to run
* Added one more cell to `CoNLL_2.ipynb` to generate the example DataFrame used in the PyData Montreal talk.
* Fixed a minor typo in a comment in `Market_Intelligence_Part1.ipynb`
* Edited the Markdown text of `Market_Intelligence_Part2.ipynb` based on edits from the Medium blog version
* Added the code-only notebooks, drawn directly from `Market_Intelligence_Part1.ipynb`, that were used in the PyData Montreal demo.
* Removed unnecessary cells from one of the Github gist notebooks for `Market_Intelligence_Part2.ipynb`
* Removed unused code from `market_intelligence.py` and synced some functions with their Github gist versions.

PR #225, opened 2021-07-09
MINOR: Address review comments from previous PR

frreiss:
Minor follow-on change to address some review comments from @BryanCutler on #208


frreiss:
Sorry about the long list of commits, this is a really tiny change mixed in with a bunch of merges.

PR #226, opened 2021-07-09
Regression tests and table style tweaks for new SpanArray render script

PokkeFe:
Added a case in the Jupyter regression test for a document with newlines.

Updated affected notebooks with new render.

Updated the table controls with SVG button icons from open-source project _Font Awesome 4_.


frreiss (approved):
LGTM

Issue #227, opened 2021-07-12
maybe_download_dataset_data() fails with example code in Read_conllu_Files.ipynb


frreiss:
Cell 3 of `Read_connlu_Files.ipynb` contains this line of code:

```python
conll_09_path = tp.io.conll.maybe_download_dataset_data(BASE_DIR, conll_09_test_data_url)
```

When I attempt to run this notebook on a freshly checked-out copy of the repository, I get the following error:
```
-------------------------------------------------------------------------
FileNotFoundError                       Traceback (most recent call last)
<ipython-input-4-2ca899c89315> in <module>
      1 # download the files if they have not already been downloaded
----> 2 conll_09_path = tp.io.conll.maybe_download_dataset_data(BASE_DIR, conll_09_test_data_url)
      3 conllu_ewt_path = tp.io.conll.maybe_download_dataset_data(BASE_DIR, ewt_dev_url)
      4 
      5 # if you already have access to the full conll 2009 dataset, name the file accordingly and uncomment this line

~/pd/tep-0_1/notebooks/../text_extensions_for_pandas/io/conll.py in maybe_download_dataset_data(target_dir, document_url, alternate_name)
   1327         alternate_name is None or not os.path.exists(full_path)
   1328     ):
-> 1329         with ZipFile(full_path, "r") as zipf:
   1330             fnames = zipf.namelist()
   1331             if alternate_name is not None and alternate_name in fnames:

~/opt/miniconda3/envs/pd/lib/python3.8/zipfile.py in __init__(self, file, mode, compression, allowZip64, compresslevel, strict_timestamps)
   1249             while True:
   1250                 try:
-> 1251                     self.fp = io.open(file, filemode)
   1252                 except OSError:
   1253                     if filemode in modeDict:

FileNotFoundError: [Errno 2] No such file or directory: 'CoNLL_u_test_inputs/CoNLL2009-ST-English-trial.zip'
```

The logic that is raising this exception appears to be incorrect. The code assumes that a file that it has not downloaded exists and attempts to unzip the file without downloading it.

The fix for this problem should include a regression test to ensure the problem does not reappear in the future.

PR #228, opened 2021-07-12
Fix bug in Maybe downoad dataset; create regression tests to verify. 

ZachEichen:
Addresses issue #227 , and adds regression tests to verify it.

### DIFF
--- a/text_extensions_for_pandas/io/conll.py
+++ b/text_extensions_for_pandas/io/conll.py
@@ -1320,13 +1320,23 @@ def maybe_download_dataset_data(
     file_name = (
         alternate_name if alternate_name is not None else document_url.split("/")[-1]
     )
-    full_path = target_dir + file_name
+    full_path = target_dir +'/' + file_name
+    # if no directory exists, create one
+    if not os.path.exists(target_dir):
+        os.mkdir(target_dir)
 
     # special logic for zip files
     if document_url.split(".")[-1] == "zip" and (
         alternate_name is None or not os.path.exists(full_path)
     ):
-        with ZipFile(full_path, "r") as zipf:
+        # if we have a zip file already, don't re-download it 
+        zipPath = target_dir +'/'+ document_url.split("/")[-1]
+        if not os.path.exists(zipPath):
+            data = requests.get(document_url)
+            open(zipPath, "wb").write(data.content)
+        
+        # if need be, extract the zipfile documents 
+        with ZipFile(zipPath, "r") as zipf:
             fnames = zipf.namelist()
             if alternate_name is not None and alternate_name in fnames:
                 zipf.extract(alternate_name, target_dir)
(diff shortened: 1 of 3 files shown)


text_extensions_for_pandas/io/conll.py
```diff
@@ -1302,31 +1302,41 @@ def maybe_download_dataset_data(
...
-
-   :param target_dir: Directory where this function should write the document
-   :param document_url: url from which to download the docuemnt. If no alternate name is specified,
-    it is assumed that the string after the last slash is the name of the file.
-   :param alternate_name: if given, the name of the file that is checked in the target directory,
-    as well as what is used to save the file if no such file is found. If a zip file is downloaded, and a file of this
-    name exists in in the archive, only it will be extracted.
-
-    :returns: the path to the file, or None if downloading was not successful
+     If the file found at the github url is not found in the target directory,
```
frreiss, line 1305:
```suggestion
     If the file found at the url is not found in the target directory,
```

text_extensions_for_pandas/io/conll.py
```diff
@@ -1302,31 +1302,41 @@ def maybe_download_dataset_data(
...
+     downloads it from the github url, and saves it to that plave in downloads.
```
frreiss, line 1306:
```suggestion
     downloads it from the url, and saves it to that place in downloads.
```

text_extensions_for_pandas/io/test_conll.py
```diff
@@ -731,6 +732,43 @@ def test_compute_accuracy(self):
...
+        self.assertTrue(os.path.isdir(ewt_dir))
+        self.assertTrue(os.path.isfile(ewt_dir + "/en_ewt-ud-dev.conllu"))
+        # test download of
+        val = maybe_download_dataset_data(ewt_dir, ewt_url, alternate_name="dev.conllu")
+        self.assertEqual(val, ewt_dir + "/dev.conllu")
+        self.assertTrue(os.path.isdir(ewt_dir))
+        self.assertTrue(os.path.isfile(ewt_dir + "/dev.conllu"))
+        # check we didn't overwrite the last file
+        self.assertTrue(os.path.isfile(ewt_dir + "/en_ewt-ud-dev.conllu"))
+
```
frreiss, line 756: You should call `maybe_download_dataset_data()` again here to make sure that the function works if the local file exists.

text_extensions_for_pandas/io/test_conll.py
```diff
@@ -731,6 +732,43 @@ def test_compute_accuracy(self):
             ),
         )
 
+    def test_maybe_download_dataset(self):
+        base_dir = "test_data/io/test_conll"
```
frreiss, line 736: Need to replace this fixed location with a temporary directory so that subsequent runs of this test won't see data from previous runs. Also make sure to delete the temp directory if the test passes.

frreiss (changes requested):
Looking good. Some minor changes requested inline.

Can you also please rerun the affected notebook and attach updated output to this PR?

text_extensions_for_pandas/io/conll.py
ZachEichen: fixed in last pull

text_extensions_for_pandas/io/test_conll.py
ZachEichen: Fixed in latest commit

text_extensions_for_pandas/io/test_conll.py
ZachEichen: fixed in latest commit.

ZachEichen:
> Can you also please rerun the affected notebook and attach updated output to this PR?
Fixed in my last commit.

frreiss (approved):

PR #230, opened 2021-07-15
Add pip to initial environment contents

frreiss:
Recent versions of Anaconda don't install `pip` by default in new environments. This causes problems when our environment setup script tries to run `pip install` from inside a newly-created environment. This PR adds `pip` to the initial environment contents to fix this problem.


BryanCutler (approved):
LGTM

PR #233, opened 2021-07-22
MINOR: add additional flexibility to a few functions within tp.io

ZachEichen:
Added flexibility to two functions within tp.io 

1. Added the option to split corpuses into documents by sentence boundary when reading in `.conllu` files using the `tp.io.conll` module.
2. Added the option to have differently named columns when using `tp.bert.align_bert_tokens_to_corpus_tokens`


frreiss (approved):
LGTM

PR #236, opened 2021-07-26
Remove version check from workaround for Pandas #42430

frreiss:
The discussion around https://github.com/pandas-dev/pandas/issues/42430 is taking longer than expected. Pandas has released a 1.3.1 release without a fix for this issue. From the direction the discussion is taking, it looks like further performance related changes are likely to introduce similar breakage in the future, even after this issue is patched.

This PR removes the Pandas version check from the workaround from this issue. That way we won't be broken if Pandas decides to put the offending performance code into another future release. Due to the other checks that are still done, the workaround is still only activated in a narrow set of circumstances even after this change.


frreiss:
Merging this PR, as this issue is blocking our CI.

PR #244, opened 2021-12-01
Minor: Add more aggregates for TensorArray

frreiss:
This PR expands the set of aggregates that can be evaluated over a `TensorArray` via `pandas.core.groupby.DataFrameGroupBy.aggregate`. All the major array->scalar reductions built into NumPy are now supported. I've streamlined the structure of the `_reduce()` method to support the longer list of aggregates without a large set of if-then statements. I also added a test case to exercise all the branches of `_reduce`.

I'm also including some .gitignore fixes that I added to a commit by accident. Don't think they'll do any harm.


BryanCutler (approved):
LGTM, thanks @frreiss !

PR #249, opened 2023-08-28
Various changes for Pandas 2.x compability.

frreiss:
This PR updates the code in the library so that regression tests, notebooks, and tutorial scenarios all work with Pandas 2.0.

I had to disable some tests relating to NumPy universal functions because upstream changes seem to have broken this functionality in a way that defies any kind of easy fix. Moving forward we'll probably want to deprecate the TensorArray class in favor of the new Arrow Tensor type plus Arrow-based arrays.

FYI @BryanCutler


BryanCutler (approved):
LGTM. Glad to see they finally merged in the tensor extension type after only a couple years! ;)

